PR #424, opened 2026-04-23
perf(metafetch): wire persistent cache into fetch path (bulk + single)

jdfalk:
## Summary

- `FetchMetadataForBook` and `bulkFetchMetadata` were calling external metadata APIs on every invocation, bypassing the `metadata_fetch_cache` KV store that `SearchMetadataForBookWithOptions` already uses
- After this fix: whichever path runs first (full library fetch, search dialog, or single-book fetch) populates the persistent cache; all subsequent calls for that `(bookID, source)` pair are instant cache hits
- Also adds a `TODO.md` note to refactor `bulkFetchMetadata` to delegate to `Service.FetchMetadataForBook` rather than duplicating the source-loop

## Test plan
- [ ] Full library bulk-fetch → check logs for `cache put` entries
- [ ] Navigate to individual book → click Fetch → logs should show `cache HIT` for sources already fetched
- [ ] Open search dialog on same book → should also show `cache HIT`
- [ ] Verify cache is invalidated when metadata is applied (existing behaviour — `InvalidateAllCachedMetadataFetchesForBook` called in apply path)


jdfalk:
@claude rebase and fix the merge conflicts.

PR #968, opened 2026-05-15
chore(gitignore): remove all .claude directory and file exclusions

jdfalk:
Removes all .claude directory and file exclusions from .gitignore to allow agent state and worktree files to be tracked.\n\nCo-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


jdfalk:
CI: Add GOEXPERIMENT=jsonv2 to the Build binary step so the built binary matches govulncheck expectations; kept govulncheck in binary mode. The .claude/.gitignore edits are on this branch because they were staged in the same feature branch — can split/revert in a follow-up if needed.

PR #1278, opened 2026-06-04
fix(ci): bump ghcommon pin to runner image with safe-ai-util-mcp symlink

jdfalk:
## Problem

`nightly-burndown.yml` was pinned to ghcommon `2e30fe8` which uses runner image `cdce859` (May 9) — before the `/opt/venv/bin` PATH setup existed. Every dispatch job failed with `safe-ai-util-mcp: executable file not found in $PATH`.

## Fix

Bump to ghcommon `8d38c68` which uses runner image `d558a337` (June 3). Paired with burndown-runner-image fix that adds the explicit symlink.

### DIFF
--- a/.github/workflows/nightly-burndown.yml
+++ b/.github/workflows/nightly-burndown.yml
@@ -1,5 +1,5 @@
 # file: .github/workflows/nightly-burndown.yml
-# version: 1.3.0
+# version: 1.4.0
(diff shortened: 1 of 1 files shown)


jdfalk:
Superseded by #1279 which pins the correct ghcommon SHA (89b4db4) using the fully-fixed runner image.

PR #1306, opened 2026-06-04
fix(security): remediate 11 pen-test findings (2 crit, 4 high, 5 med)

jdfalk:
Remediates the findings from the 2026-06-04 local penetration test. **10 fixed + 1 documented as accepted risk; 1 (MED-5) deferred pending repro.** One commit per finding for easy review/revert.

## Critical
- **CRIT-1** — Stop logging raw `abbs_`/`abk_` tokens at INFO. Bootstrap token is read from the `0600` `.bootstrap-token` file; read-only key written to a new `0600` `.readonly-key` file. Logs now carry only path + expiry. **The `server-bootstrap` skill is updated** to read the token file over SSH (it previously scraped journalctl — this fix would have broken our own bootstrap flow otherwise).
- **CRIT-2** — Stop returning the raw `*database.User` (with `password_hash`) from accept-invite + the latent deactivate/reactivate handlers. All return the safe `AuthUserResponse` shape. Regression tests assert no hash field appears.

## High
- **HIGH-1** — pprof listener is opt-in via `ABK_PPROF_ADDR` even in the `pprof` build (was an auto-started unauthenticated :6060 listener).
- **HIGH-2** — `router.SetTrustedProxies(nil)` so `X-Forwarded-For` can't spoof `ClientIP` and bypass per-IP limiters. (Deployment is direct-connect; comment documents the CIDR-allowlist path if a proxy is ever added.)
- **HIGH-3** — Replace the weaponizable hard account lockout with a **per-IP throttle** (keyed on attacker source, now spoof-resistant via HIGH-2) + a **soft per-account progressive delay** that never hard-locks the victim.
- **HIGH-4a** — Bootstrap exchange returns **401 not 500** once the one-time token is consumed, via a new `database.ErrSettingNotFound` sentinel (affects the canonical PebbleDB backend too).
- **HIGH-4b** — SQLite RBAC **fails loudly** (`ErrSQLiteRBACUnsupported`) instead of silently granting empty perms / 403-ing every request.

## Medium
- **MED-2** — `/api/events` SSE gated behind auth (cookie-based; browser UI unaffected).
- **MED-3** — Route-level `PermSettingsManage` guard on `POST /maintenance/jobs/:job_id`.
- **MED-4** — Invite `created_by_user_id` populated for the audit trail.
- **MED-1** — `/metrics` left unauthenticated **by maintainer decision** (accepted risk; scrapers send no auth, data is operational metrics). Rationale documented in code.
- **MED-5** — accept-invite EOF under HTTP/2: **deferred** (TODO.md). Not reproducible statically and not accept-invite-specific in code; leading hypothesis is the `/api/*`→`/api/v1/*` 301 dropping the POST body. Does not block the other fixes.

## Decisions taken (confirmed with maintainer)
HIGH-3 → per-IP throttle + soft counter · MED-1 → leave `/metrics` open · MED-2 → gate with cookie auth · HIGH-4b → fail loudly (don't implement deprecated-backend RBAC).

## Testing
- New regression tests: password-hash-leak (accept-invite/deactivate/reactivate), consumed-token→401, SQLite RBAC fail-loud, per-IP throttle, soft-counter-doesn't-lock, different-IP-not-throttled.
- `go build ./...`, `go vet ./...` clean. Changed packages (`server`, `handlers`, `middleware`, `auth`, `database` role/setting subset) all green.
- Pre-existing flaky `panic: pebble: closed` in `internal/database` MemStore warmup is unrelated (fails on `main` too); see notes.

### DIFF
--- a/.claude/skills/server-bootstrap/SKILL.md
+++ b/.claude/skills/server-bootstrap/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: server-bootstrap
-description: Initialize server authentication and retrieve API key. SSH to the audiobook-organizer server, restart the service, extract the bootstrap token from journalctl, exchange it for an API key via POST /api/v1/auth/bootstrap, and write the key to .claude/.api-token (shared across worktrees, auto-cleanup after 8 hours). Use when starting fresh or when the API key has expired.
+description: Initialize server authentication and retrieve API key. SSH to the audiobook-organizer server, restart the service, read the bootstrap token from the .bootstrap-token file (no longer logged in plaintext — pen-test CRIT-1), exchange it for an API key via POST /api/v1/auth/bootstrap, and write the key to .claude/.api-token (shared across worktrees, auto-cleanup after 8 hours). Use when starting fresh or when the API key has expired.
 ---
 
 # Server Bootstrap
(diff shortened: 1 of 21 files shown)


jdfalk:
### CI baseline + verification notes

**CI red is pre-existing, not from this PR.** `main`'s `Continuous Integration` workflow already fails at its base commit (`4d8b3fe7`, which this branch is built on): the `Go Tests (short, race)` job fails on a **DATA RACE in `TestServerStartGracefulShutdown`** plus the `panic: pebble: closed` memdb-warmup flake. `Go Vet & Build`, mock freshness, and all frontend jobs are green on main. This PR adds no goroutines or shared state to server startup, so it inherits that same red and introduces **zero new failures**. Locally: `go build ./...` + `go vet ./...` clean; the changed packages (`server`, `handlers`, `middleware`, `auth`, and the `database` role/setting tests) all pass.

**CRIT-2 class confirmed closed.** Swept all `internal/server` non-test code for user-bearing responses: every one uses `BuildAuthUserResponse`/`buildAuthUserResponse` or an explicit field-selected `gin.H` (e.g. `auth_temp_login.go` returns only `id`+`username`). No raw `*database.User` is serialized anywhere. (Note: a structural `json:"-"` on `PasswordHash` was deliberately **not** used — PebbleStore persists the same struct via `json.Marshal`, so the tag would drop the hash from storage and break auth. Per-call-site is the correct fix.)

**Two non-blocking notes for reviewers:**
- `InitStartupReadOnlyKey`'s new `.readonly-key` write path is untested; `filepath.Dir("")` → `"."` would write the key to CWD if `DatabasePath` were ever empty (prod always sets it, so low risk).
- `acctFails`/`ipFails` maps evict only on access (no background TTL) — same shape as the old `lockout` map they replace. Fine to leave; could add idle eviction later for parity with the IP rate-limiter.

jdfalk:
CI re-trigger after lgtm annotation push

PR #1368, opened 2026-06-10
fix(lifecycle): join background goroutines on shutdown; fix scanner config race (fable5 T027)

jdfalk:
## Summary

- **Part A (dedup Engine)**: `bgWg sync.WaitGroup` added; `bgWg.Add(1)` called under `bgMu` before goroutine launch; `Stop()` now cancels context and waits with a bounded 5 s timeout (per-engine `stopTimeout_` field, overridable in tests); `WARN` on expiry, no hang. Double-`Stop()` is idempotent. 4 new lifecycle tests with `-race`.
- **Part B (server namedWaitGroup)**: New `bg_wg.go` adds `namedWaitGroup` wrapping `sync.WaitGroup` with live goroutine name tracking; 8 server lifecycle goroutines now use named `Add/Done`; shutdown `WARN` log includes `"still_running": [...]` for operator diagnostics.
- **Part C (scanner config race)**: Root-cause fix for the CI data race `TestServerStartGracefulShutdown` — `saveBookToDatabase` now (a) accepts `ctx context.Context`, (b) checks `ctx.Err()` at entry, and (c) snapshots `config.AppConfig.RootDir` into a local variable immediately, preventing concurrent reads of the global after test cleanup writes it. All 4 `saveBook` call sites and all test-seam overrides updated.

## Root-cause writeup

The CI race manifested as:

```
write by goroutine N (test cleanup): config.AppConfig.RootDir = ...
read by goroutine M (scanner worker): config.AppConfig.RootDir (saveBookToDatabase:1700+)
```

Workers launched by `ProcessBooksParallel` entered `saveBookToDatabase` and read `config.AppConfig.RootDir` multiple times (lines 1700, 1789-1791, 1801-1802, 1874-1875). If the calling test cancelled its context and restored `config.AppConfig` before those goroutines finished, the race fired.

**Why the existing registry drain fix (`be8aae96`) was not sufficient**: the drain ensured operation *handles* outlived the goroutines, but scan-op goroutines don't check context between the wg.Wait semaphore release and the `saveBookToDatabase` call — they enter the function first, then read the global.

**Fix**: snapshot `config.AppConfig.RootDir` at function entry (one read, under no lock, before any goroutines are in-flight). Combined with the ctx early-exit, workers that run past context cancellation still use the snapshotted value, not the post-cleanup value.

## Acceptance checklist

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean  
- [x] `go test ./internal/dedup/... -race -count=1` — PASS
- [x] `go test ./internal/scanner/... -race -count=1` — PASS (includes 2 new regression tests)
- [ ] `go test ./internal/server/... -race -count=1` — PASS (handler sub-packages all pass; `TestProp_ChromemMatchesSqlite` in root `internal/server` is a **pre-existing** failure that also fails on main at `26f34dc3` — see evidence below)
- [ ] `make test` — pending (blocked on same pre-existing failure)

### Pre-existing failure evidence

`TestProp_ChromemMatchesSqlite` fails on the worktree base commit (`26f34dc3`, before any T027 changes):
```
dedup_engine_prop_test.go:386: sqlite→chromem overlap too low: 0 of 1 matched (chromem set=0)
FAIL github.com/falkcorp/audiobook-organizer/internal/server 1.669s
```
This is a property-based chromem vector test that has a non-deterministic failure mode unrelated to this PR.

---

COMPLETED: 3 — Part A dedup lifecycle hardening, Part B namedWaitGroup instrumentation, Part C scanner config race fix
REMAINING: 0
BLOCKED: 0

### DIFF
--- a/internal/dedup/engine.go
+++ b/internal/dedup/engine.go
@@ -1,5 +1,5 @@
 // file: internal/dedup/engine.go
-// version: 1.22.0
+// version: 1.23.0
 // guid: 8f3a1c6e-d472-4b9a-a5e1-7c2d9f0b3e84
 // last-edited: 2026-05-18
 
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/falkcorp/audiobook-organizer/internal/ai"
 	"github.com/falkcorp/audiobook-organizer/internal/ai/aijobs"
@@ -83,9 +84,17 @@ type Engine struct {
 	// engine doesn't need a New-time ctx and Stop can cleanly cancel
 	// outstanding work. Guarded by bgMu (RWMutex — readers are subscriber
 	// goroutines that snapshot the ctx; writers are PostInit and Stop).
+	//
+	// bgWg tracks goroutines started under bgCtx so Stop can join them and
+	// confirm the Pebble read is fully finished before the store closes.
 	bgCtx    context.Context
 	bgCancel context.CancelFunc
 	bgMu     sync.RWMutex
+	bgWg     sync.WaitGroup
+
+	// stopTimeout overrides the default join timeout in Stop.  Zero means use
+	// defaultHydrationStopTimeout.  Only set in tests.
+	stopTimeout_ time.Duration
 }
 
 // NewEngine creates a Engine with sensible defaults.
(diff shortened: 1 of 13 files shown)


jdfalk:
Coordinator note: the race-job failure on this run is a DIFFERENT instance of the same family — write at internal/config/update_service.go:130 (config update service mutating global AppConfig, triggered from service_layer_test.go:993) racing a background reader. Pre-existing on main, not introduced here. This PR's scanner-snapshot + joins fix the instance that was flaking CI; the structural cure (synchronized AppConfig — atomic.Pointer[Config] or accessor with RWMutex) is being queued as fable5 follow-up task T028.

PR #1463, opened 2026-06-15
feat(plugin): emergency-console agent

jdfalk:
## Summary

- Adds `.claude-plugin/agents/emergency-console.md` — a red, Bash-only plugin agent
- Single purpose: dry-run → confirm → restart all other Claude Code sessions via `restart-other-claudes` skill
- Triggers on: "restart all claudes", "fire", "what would get restarted"

## Companion files (not in this repo)

These are installed globally and not committed here:

- `~/.claude/emergency-console/CLAUDE.md` — session spec for top-level `claude-loop` mode
- `~/.local/bin/emergency-console` — launcher: creates tmux session, runs `claude-loop` in the console dir

## Usage

**As a plugin subagent** (from any Claude session): just say "restart all claudes" or "fire"

**As a persistent standalone session:**
```bash
emergency-console   # launches in tmux, survives disconnects, resumes via --continue
```

### DIFF
--- a/.claude-plugin/agents/emergency-console.md
+++ b/.claude-plugin/agents/emergency-console.md
@@ -1,0 +1,66 @@
+---
+name: emergency-console
+description: |
+  Use this agent when you need to kill and restart other Claude Code sessions on this machine.
+  This agent is the emergency reset button -- signals all other claude processes and lets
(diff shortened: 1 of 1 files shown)


jdfalk:
Wrong repo — moving to standalone plugin

PR #1953, opened 2026-07-14
feat(regroup): review-queue APPLY path for confident holds (PR-B2)

jdfalk:
## What

Turns the review queue's **Approve** from a status label-flip into a **real merge** for the two *confident* regroup kinds — closing the loop opened by B1 (the dry-run producer, #1949). Approving a hold dispatches on its `Kind` to a registered `ApplyFunc`; on a nil error the handler transitions the item to `applied`.

| Kind | Prod holds | Apply action |
|------|-----------:|--------------|
| `regroup.multidisc` | 196 | `CombineBooks(members, primary, nil)` — collapse N single-file books into 1 |
| `regroup.version-group` | 1 | shared `VersionGroupID` + one primary via re-fetch-and-patch `UpdateBook` |
| `regroup.anthology` | 8 | **handler-less** — approve → `approved`, never `applied` |
| `regroup.ambiguous` | 230 | **handler-less** — needs human sub-decisions |

## Data-loss safety (this repo's dominant incident class)

- **multidisc**: the **nil override** is load-bearing — the only `UpdateBook`-on-survivor path in `merge.Service` is gated behind a non-nil override, so with nil the survivor row is never rewritten and its `AcoustIDFingerprint` / Author / Series survive. Absorbed files move by file ID (fingerprints ride along); absorbed rows are hard-deleted (files preserved on survivor → no audio loss).
- **version-group**: re-fetch-and-patch (`GetBookByID` full row → set only `VersionGroupID`+`IsPrimaryVersion` → `UpdateBook`). **NOT `MergeBooks`** — that soft-deletes the loser, but a version group must keep **both** editions visible (locked decision #8).
- Tests assert these invariants via `dbtest.AssertStoreInvariants` — the same drift-proof guard as the Jul-13 data-loss fixes.

## Design notes

- Survivor selection is deterministic (smallest ULID = earliest-created), so retries are stable.
- Apply is **retry-tolerant**: fewer than two surviving members is an idempotent no-op, so a double-approve or re-approve after a partial failure never errors.
- Apply closures are built at wire time (where the merge service + store are in scope) and registered via `reviewH.RegisterApplyHandler`, keeping the `reviewhandler` package generic.

## ⚠️ Not auto-applied — gated

This PR only **arms** the button. Every collapse remains a **human-approved hold**; nothing auto-applies. **Merging + deploying is the gated prod-apply decision** — hold until explicitly approved.

## Test

- `go test ./internal/plugins/maintenance/ -run 'TestApply|TestPickPrimary' -race` — green (collapse + preserve-data invariants, version-group primary/both-visible, already-applied no-op, bad-payload error).
- Full touched-package suites (`maintenance`, `server`, `merge`, `review` handler) — green. gofmt/vet/staticcheck clean.

Plan: `docs/plans/2026-07-13-review-queue-and-regroup.md` (PR-B2).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 <!-- file: CHANGELOG.md -->
-<!-- version: 3.157.2 -->
+<!-- version: 3.158.0 -->
 <!-- guid: 8c5a02ad-7cfe-4c6d-a4b7-3d5f92daabc1 -->
 <!-- last-edited: 2026-07-16 -->
 
@@ -82,8 +82,13 @@ success the item transitions to `applied`.
   `internal/server/wire_handlers.go`. Tests assert the data-loss invariants (survivor
   keeps fingerprint + author; version-group members keep theirs) via
   `dbtest.AssertStoreInvariants` — the same drift-proof guard as the Jul-13 fixes.
-- **Not auto-applied.** This PR only *arms* the button; every collapse remains a
-  human-approved hold. Merging + deploying is the gated prod-apply decision.
+- **Global apply "off switch" (`config.review_apply_enabled`, default OFF).** Even with
+  this code deployed, approving a hold records the decision but NEVER runs the merge
+  while the switch is off — everything stays visible in the review pane. The gate lives
+  in the review handler (producer-agnostic) and is read at approve time. Flip to `true`
+  only when you want approvals to actually apply. New tests cover both switch states.
+- **Not auto-applied.** This PR only *arms* the button, and even then only when the
+  switch is on; every collapse remains a human-approved hold.
 
 #### July 13, 2026 - fix(regroup): anthology counts distinct works + fold in original iTunes album path (B1 tuning)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 <!-- file: CHANGELOG.md -->
-<!-- version: 3.157.1 -->
+<!-- version: 3.157.2 -->
 <!-- guid: 8c5a02ad-7cfe-4c6d-a4b7-3d5f92daabc1 -->
 <!-- last-edited: 2026-07-16 -->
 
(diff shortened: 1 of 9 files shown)


jdfalk:
### Reviewer notes — deferred decisions (surfaced, not oversights)

Three deliberate choices the tests pass *around* — flagging so they're seen, not silent:

1. **Multidisc keeps the survivor's raw track title, not the cleaned `SurvivorTitle`.** `CombineBooks(..., nil)` uses a nil override (the data-loss-safe choice → survivor row never rewritten), so the collapsed book inherits the smallest-ULID member's existing title (e.g. `01 - The Ambassador's Mission`). The payload's `SurvivorTitle` is currently **unused on apply**. A follow-up can pass a title-only override after re-fetching the survivor (the same safe patch `CombineBooks` already uses) if a cleaned title is wanted.
2. **Version-group primary = smallest ULID may not be the Unabridged edition.** The plan prefers Unabridged=primary; smallest-ULID (earliest-imported) has no correlation with abridged/unabridged. Only **1 hold** in the whole library today, so low stakes — but the "right" primary needs a signal the payload doesn't carry.
3. **Survivor selection = smallest ULID** (both kinds) — deterministic and retry-stable, but if a richer survivor signal is wanted, B1 could record an explicit `SurvivorBookID`. The apply path is forward-compatible either way.

**Executive summary deliberately deferred:** this qualifies on "wide blast radius," but nothing is user-facing until it's merged + a human approves a hold on prod. The exec summary belongs when the apply actually ships/runs — recording this as a *decision*, not an omission.

**Draft on purpose:** this arms the Approve button to run real `CombineBooks` merges. Kept as a draft so the admin-merge-on-green automation can't merge it — mark ready only with an explicit prod-apply decision.

jdfalk:
### Added: global apply "off switch" (per feedback — no auto-merge until the big switch)

New commit `3b8cff5`. Even with this code deployed, **nothing merges when a human clicks Approve until the switch is on** — everything stays visible in the review pane.

- **`config.review_apply_enabled`** (default **OFF**). When off, approving a hold records it as `approved` but never runs the apply handler; the response note says review-only mode. When on, approve executes `CombineBooks`/version-group and marks `applied`.
- Gate lives in the review handler (`applyEnabled func() bool`, read at approve time), so it's producer-agnostic and a config change takes effect without re-wiring. A nil gate is fail-safe OFF.
- Apply handlers are still *registered* when off, so flipping the switch on is the only step needed to enable applies.
- Tests: `TestApproveReviewItem_ApplyGateOff_ApprovesNotApplies` (handler registered, switch off → approved, handler NOT run) and `_ApplyGateOn_Applies`.

This reverses the earlier "dropped the confident-auto-apply toggle" note — the explicit requirement is: see everything in the review pane, nothing auto-merges until the switch is flipped. **Two locks now:** PR stays draft, AND the switch defaults off even after merge.

PR #1966, opened 2026-07-16
fix(audiobooks): editing a book's author by ID persisted a stale author name

jdfalk:
## Problem

`UpdateAudiobook` (`internal/audiobooks/service_mutation.go`) changes a book's `AuthorID`/`SeriesID` — reachable via the `author_id`/`series_id` API fields (`update_service.go:86`) — but left the **embedded, denormalized `Author`/`Series` display object** pointing at the *old* record.

- `GetBookByID` loads the book with its old embedded `Author`.
- Setting `currentBook.AuthorID` changes the FK but not `currentBook.Author`.
- `UpdateBook`'s preserve-on-nil guard (`if book.Author == nil`) can't fix a **stale non-nil** struct, so the stale name is marshaled into the stored blob.
- The response-enrichment block (L369-374) runs *after* `UpdateBook` returns the same pointer, so it fixes only the in-memory reply — not the persisted row.
- Every read path — `resolveAuthorAndSeriesNames` (helpers.go) and `EnrichAudiobooksWithNames` (service_query.go) — **prefers the embedded object when non-nil**, only falling back to `GetAuthorByID` when it's `nil`.

Net effect: after changing a book's author (or series) by ID, the book's FK points at the new author while every screen displays the **old** author name — indefinitely, with no error. This is the inverse of the 2026-07-13 blank-on-nil author/series data-loss fix (that one handled nil; this one handles stale-non-nil).

## Fix

Sync the embedded `Author`/`Series` display object to the resolved ID **before** persisting, honoring `UpdateBook`'s documented "a write that changes the author must set BOTH the ID and a fresh object" contract. Reuses the exact construction from the existing post-persist enrichment (so the multi-author "A & B" combined-display case is preserved). Because it runs on every edit, it also self-heals any pre-existing mismatch.

## Test

`TestUpdateAudiobook_SyncsDenormalizedAuthorOnIDChange` seeds a book via a real `PebbleStore`, changes the author via `author_id` only, then **re-reads the row straight from the store** (not the returned pointer, which the response-enrichment fixes regardless) and asserts the persisted `Author.ID` **and** `Author.Name` track the new author. Verified it FAILS without the fix (persisted `Author.ID` stays the old author = 1) and passes with it.

- `go build ./...` clean
- `go vet ./internal/audiobooks/` clean
- `go test ./internal/audiobooks/ -race` green (48.9s)

Docs: CHANGELOG + TODO (`UPDATEBOOK-STALE-DENORMALIZED-AUTHOR`) + executive summary (`docs/executive-summaries/2026-07-16-stale-author-on-id-edit-executive-summary.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 <!-- file: CHANGELOG.md -->
-<!-- version: 3.164.0 -->
+<!-- version: 3.165.0 -->
 <!-- guid: 8c5a02ad-7cfe-4c6d-a4b7-3d5f92daabc1 -->
 <!-- last-edited: 2026-07-16 -->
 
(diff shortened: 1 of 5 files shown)


jdfalk:

PR #1967, opened 2026-07-16
fix(server): unify destructive-reset authz on RequireAdmin + confirm token on /system/reset

jdfalk:
Closes the **RESET-AUTHZ-CONSISTENCY** TODO item (API authz bug hunt, 2026-07-16, LOW).

## Two gaps on the full-wipe endpoints

**1. Inconsistent authz.** `/system/reset` and `/system/factory-reset` gated on `PermSettingsManage`, while the sibling destructive endpoint `/maintenance/wipe` (and `/admin/*`, `/cache/stats/keys`) gate on `RequireAdmin()`. In the default roles only admin has `settings.manage`, so there's no live escalation — but a **custom non-admin role granted `settings.manage`** could wipe the entire library via reset. Both reset endpoints now go through the same admin-only subgroup as `/maintenance/wipe`, using the established pattern:

```go
adminOnly := protected.Group("")
adminOnly.Use(servermiddleware.RequireAdmin())
{
    adminOnly.POST("/system/reset", systemH.ResetSystem)
    adminOnly.POST("/system/factory-reset", systemH.FactoryReset)
}
```

**2. Missing confirm token.** `ResetSystem` performs a full DB wipe (books, authors, series, settings) but accepted a **bare POST**, while the *more* destructive `FactoryReset` already required `{"confirm":"RESET"}`. `ResetSystem` now requires the same token — a mis-fired or accidental request can't silently erase the database.

## Behavior note

When `EnableAuth=false` the `authMiddleware` is a passthrough that injects no user, so `RequireAdmin()` 401s — meaning these two endpoints become unreachable in auth-disabled dev mode, exactly like `/maintenance/wipe` already is. That's the intended, safer posture for a destructive full-wipe op and the point of unifying on one model.

## Test

- `TestResetSystem_RequiresConfirm` (new): a wrong/absent token → **400**, and the mock store asserts `Reset()` is **never called**. Fails without the confirm gate (previously 200 + DB wiped).
- `TestResetSystem_OK` / `TestResetSystem_DBError` updated to send the token.
- `go build ./...`, `go vet ./internal/server/...` clean; `go test ./internal/server/handlers/system/` green.

Docs: CHANGELOG + TODO updated. No executive summary — LOW-severity defense-in-depth hardening, single-PR, no observed data loss.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 <!-- file: CHANGELOG.md -->
-<!-- version: 3.165.0 -->
+<!-- version: 3.166.0 -->
 <!-- guid: 8c5a02ad-7cfe-4c6d-a4b7-3d5f92daabc1 -->
 <!-- last-edited: 2026-07-16 -->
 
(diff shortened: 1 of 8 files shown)


jdfalk:
### Follow-up: all `/system/reset` callers updated + E2E note

A review pass found bare-POST callers of `/system/reset` beyond the `handlers/system` unit tests that the confirm-token change would have 400'd. All fixed in the latest commit:

- `internal/server/reset_handler_test.go` (4 cases) and `internal/server/handlers_unit_test.go` (2 cases) now send `{"confirm":"RESET"}`.
- `web/tests/e2e/utils/setup-modes.ts` (`resetToFactoryDefaults`) now sends the token.

**Frontend:** the production UI never calls bare `/system/reset` — it only calls `/system/factory-reset`, which already sends the token. So no user-facing flow changes from the confirm token.

**E2E / auth-disabled note for the reviewer:** because `RequireAdmin()` requires an authenticated admin user (no `EnableAuth=false` passthrough, unlike the old `s.perm`), `/system/reset` and `/system/factory-reset` become unreachable when auth is disabled — exactly matching the sibling `/maintenance/wipe`, which already behaves this way. The E2E `resetToFactoryDefaults` helper is failure-tolerant by design (its own spec is "handles missing endpoint gracefully"), so it degrades gracefully rather than erroring. This is the intended unification per the TODO; flagging it in case the E2E environment relied on the old auth-disabled passthrough for reset.

jdfalk:

PR #1968, opened 2026-07-16
perf(reconcile): parallelize the last two serial whole-library loops

jdfalk:
Closes the **RECONCILE-SERIAL-LOOPS** TODO item — the follow-up to #1960 (`hashFilesConcurrent`). The two remaining single-core hotspots in `internal/reconcile/reconcile.go` now run across a bounded `errgroup` pool sized to `runtime.NumCPU()`.

## What changed

**1. `BuildReconcilePreviewWithProgress` path-check** (the per-book `os.Stat` over the whole library):
- `knownPaths` is a cheap in-memory set with no I/O → built in a serial pre-pass.
- The expensive `os.Stat` calls run concurrently, each recording into its own index slot (`brokenFlag[i]`).
- The ordered `brokenBooks` / `BrokenRecords` fold runs serially afterward → **byte-for-byte identical** to the former sequential build.

**2. `FindBrokenSegmentBooks`** (directory stat + `GetBookFiles` + per-segment stat storm + non-dry-run hydrate/`UpdateBook`):
- The **outer** per-book loop is sharded across workers.
- **Write safety:** work partitions by book index, so each book's DB row is touched by exactly one worker; the per-book `UpdateBook` writes are to disjoint keys and never race. They only flip `LibraryState`/`MarkedForDeletion` — no secondary-indexed field — so there's no cross-book index contention either.
- `Details` lands in per-index slots and folds in book order; `BooksChecked`/`BrokenBooks`/`MarkedForReview` are atomic counters assigned once at the end.

Output is unchanged: same broken records, same order, same counts. Follows the `hashFilesConcurrent` pattern and the CLAUDE.md concurrency mandate (partition order-sensitive shared state, don't naive-fan-out).

## Tests

Adds the package's first behavioral tests (`reconcile_parallel_test.go`), run under `-race` via a small concurrency-safe fake `Store`:
- `TestFindBrokenSegmentBooks_ParallelOrderAndCounts` — non-contiguous broken indices `{1,3,4,6}`; asserts `Details` order matches book order (a scrambled fold would fail), exact counters, dry-run writes nothing, and the real run marks exactly the broken books `needs_review`.
- `TestBuildReconcilePreview_ParallelBrokenOrder` — non-contiguous broken `{0,2,5}`; asserts `BrokenRecords` order matches book order.

`go build ./...`, `go vet ./internal/reconcile/` clean; `go test ./internal/reconcile/ -race` green.

Docs: CHANGELOG + TODO updated. No executive summary — pure performance change with identical output (matching how the sibling #1960 was handled).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 <!-- file: CHANGELOG.md -->
-<!-- version: 3.166.0 -->
+<!-- version: 3.167.0 -->
 <!-- guid: 8c5a02ad-7cfe-4c6d-a4b7-3d5f92daabc1 -->
 <!-- last-edited: 2026-07-16 -->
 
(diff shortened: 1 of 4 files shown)


jdfalk:

PR #1969, opened 2026-07-16
fix(tests): update stale GetAllBooksCore mock expectations after the unbounded-fetch change (unblocks main)

jdfalk:
## Main's `Go Tests (short, race)` gate is red — this fixes it

PRs #1961/#1962 (the whole-library truncation fix, already merged + deployed) changed `internal/itunes/service/importer.go` and `path_reconcile.go` to call `GetAllBooksCore(0, 0)` (unbounded) instead of the former fixed limits `(100000, 0)` / `(10000, 0)`. The **source** change is correct — but the mock-based tests in the same package still asserted the *old* exact args.

testify mocks match arguments exactly, so every affected test now fails with `Unexpected Method Call: GetAllBooksCore ... Diff: 0: FAIL: (int=0) != (int=100000)`. This is reproducible on `main` (19e129d4) with `go test ./internal/itunes/service/ -short -race` — it is not caused by any in-flight branch; it landed with #1962 whose own tests weren't updated in lockstep.

## Fix (test-only)

Update the stale expectations to `(0, 0)` to match the merged source:
- `importer_mock_test.go`, `importer_error_paths_test.go`, `importer_organize_parallel_test.go`, `path_reconcile_test.go`: `100000 → 0` (11 expectations)
- `importer_enrich_parallel_test.go`: `10000 → 0` (2 expectations)

The two `GetAllBooksCore(100, 0)` expectations are **left as-is** — they correctly match the paged loop at `importer.go:784`.

Verified: `go test ./internal/itunes/service/ -short -race` now passes (was the only package failing on main's CI). No production behavior change.

> Note: this red gate is why sibling PRs #1966 / #1967 / #1968 also show a failing `Go Tests (short, race)` — they branch off the same red `main`. Merging this first (then rebasing them) turns all four green.

### DIFF
--- a/internal/itunes/service/importer_enrich_parallel_test.go
+++ b/internal/itunes/service/importer_enrich_parallel_test.go
@@ -1,7 +1,7 @@
 // file: internal/itunes/service/importer_enrich_parallel_test.go
-// version: 1.1.0
+// version: 1.1.1
 // guid: 8d3a5b1e-6f2c-4a97-9e1d-3c7b8f4a2d6e
-// last-edited: 2026-07-07
+// last-edited: 2026-07-16
 
 package itunesservice
 
@@ -102,7 +102,7 @@ func buildEnrichFixture(t *testing.T, n int) ([]database.Book, *dbmocks.MockStor
 	}
 
 	m := dbmocks.NewMockStore(t)
-	m.EXPECT().GetAllBooksCore(10000, 0).Return(cores, nil)
+	m.EXPECT().GetAllBooksCore(0, 0).Return(cores, nil)
 	for i := range books {
 		id := books[i].ID
 		m.EXPECT().GetBookAuthors(id).Return(nil, nil).Maybe()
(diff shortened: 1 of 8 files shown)


jdfalk:
### Scope expanded: 3 packages, not just itunes/service

Running the affected packages locally on `main` (19e129d4) showed the stale-mock breakage from #1961/#1962 hit **three** packages, all now fixed in this PR:

- `internal/itunes/service` — `100000`/`10000 → 0` (13 expectations across 5 test files)
- `internal/server/handlers/operations` — `10000`/`100000 → 0` (`TestOptimizeDatabase_NoBooks`, `TestAuditFileConsistency_Empty`)
- `internal/server/handlers/metadata` — `1_000_000 → 0` (`TestHandleBulkWriteBack_DryRun`, `TestHandleBulkWriteBack_Enqueue202`)

`internal/diagnostics` uses a genuine paged `GetAllBooksCore(pageSize, offset)` loop and was already correct — left untouched. All three packages pass `-short -race` locally after the fix. The two legitimate `GetAllBooksCore(100, 0)` expectations (paged loop at `importer.go:784`) are preserved.

Test-only change; no production behavior change. This is the pre-existing `main` breakage that's also reddening sibling PRs #1966/#1967/#1968.

PR #2002, opened 2026-07-18
fix(registry): publish op.terminal SSE + reporter/scanner hygiene (T06+T08)

jdfalk:
Combined T06 + T08 from `docs/agent-tasks/error-correction-2026-07/TASKS.md` (both edit `internal/operations/registry/worker.go`, so they cannot be split across branches).

## R-1 (T06) — `op.terminal` SSE publisher
The event contract existed only on the frontend (`web/src/stores/useOperationsStore.ts`, `web/src/services/api.ts`) — **no backend code published it**, so completed ops lingered as phantom "running" in the operations bell until a manual refresh. Added `Registry.publishOpTerminal` (fans out `{op_id, def_id, status}` via the existing bus; nil-safe; no direct SSE-hub import) and call it on every terminal transition:
- completed / failed / canceled / timeout (in-process + subprocess paths)
- abandonment (`interrupted_*`)
- force-drop (infinite-restart)
- canceled-before-start (queued-in-worker-channel)
- canceled-while-queued (`Cancel()` of a purely-queued op — the path with no worker to publish on its behalf)

## R-3 (T08) — abandoned-reporter log buffer
After a run is abandoned its reporter's `flushLoop` has already exited (runCtx canceled) but the wedged goroutine kept calling `Log`, growing `logBuf` without bound while every line was unrecoverable anyway. Now: buffer capped at 1,000 entries (drop-oldest + dropped-counter) **and** the worker flips a terminal flag (`markTerminal`) on the abandonment paths so subsequent `Log` calls are cheap no-ops.

## R-7 (T08) — dead scan-checkpoint code
Deleted `ScanService.PerformScanWithID` (zero callers) and its `operations.SaveParams(ScanParams)`/`ClearState` calls. `LoadParams[ScanParams]` had zero callers and `library.scan` uses `ResumePolicy=ResumeDrop`, so an interrupted scan restarts from scratch (idempotent re-walk) — it never resumed from the checkpoint that was being written. Comment left on `PerformScan` explaining why. (Unused `internal/operations` import removed.)

## P-2 (T08) — RunItems backwards progress
Parallel `RunItems` reported the item **index** (`i+1`), so a high-index item finishing early reported a high value while few items had completed — the bar jumped backwards. Now reports an atomic completion counter (monotonic in both sequential and parallel modes).

## Tests
- `terminal_events_test.go` — all terminal paths fire `op.terminal` via a mutex-protected recording bus (completed, failed, canceled-running, abandoned, force-dropped, canceled-queued). State set before `Start` to avoid enqueue-vs-setup races.
- `reporter_terminal_test.go` — white-box (internal package): buffer cap drops oldest + counts, `Log` no-ops after `markTerminal`, idempotent.
- `run_items_p2_test.go` — deterministic order test: with item 0 blocked, max reported progress stays ≤ N-1 (fails under the index-based bug), final == N.

## Verification
- `go test ./internal/operations/registry/... -race -count=3` → ok (196.9s)
- `go test ./internal/scanner/... -race -count=1` → ok
- `go build ./...` → clean

Live SSE curl verification: see PR comment (attempted per T06 acceptance).

https://claude.ai/code/session_01WdKeHNKm5K6oyop4bYNd65

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,42 @@
 <!-- file: CHANGELOG.md -->
-<!-- version: 3.174.0 -->
+<!-- version: 3.175.0 -->
 <!-- guid: 8c5a02ad-7cfe-4c6d-a4b7-3d5f92daabc1 -->
-<!-- last-edited: 2026-07-17 -->
+<!-- last-edited: 2026-07-18 -->
 
 # Changelog
 
 ## [Unreleased]
 
 ### Features & Fixes
+
+#### July 18, 2026 - registry SSE + reporter/scanner hygiene (T06 + T08)
+
+- **R-1 (op.terminal SSE, T06):** the backend now publishes an `op.terminal`
+  event on every terminal transition — completed, failed, canceled, timeout,
+  abandonment (`interrupted_*`), force-drop, canceled-before-start, and
+  canceled-while-queued. Previously the event contract existed only on the
+  frontend, so finished operations lingered as phantom "running" entries in the
+  operations bell until a manual refresh. New helper
+  `Registry.publishOpTerminal` fans out `{op_id, def_id, status}` via the
+  existing bus (nil-safe; no direct SSE-hub import). Covered by
+  `terminal_events_test.go`.
+- **R-3 (abandoned-reporter log buffer, T08):** after a run is abandoned its
+  reporter's flush goroutine has already exited, but the wedged goroutine kept
+  calling `Log`, growing `logBuf` without bound and losing every line anyway.
+  The buffer is now capped at 1,000 entries (drop-oldest with a dropped-counter)
+  and the worker flips a terminal flag (`markTerminal`) on the abandonment paths
+  so subsequent `Log` calls are cheap no-ops. Covered by
+  `reporter_terminal_test.go`.
+- **R-7 (dead scan-checkpoint code, T08):** removed the unused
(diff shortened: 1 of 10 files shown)


jdfalk:
### Live SSE verification note (T06 acceptance)

**Not run as a live curl in this environment**, transparently: a local \`./audiobook-organizer serve\` needs a configured data-dir + library path + an auth bootstrap (the operations SSE route \`/api/v1/operations/events\` is gated by \`PermLibraryView\`), and standing that up here risks opening production data. Rather than claim a manual check I didn't perform, here is what actually covers the six paths:

- **`TestTerminalEvent_Completed`** drives the **real** `Registry` + **real** worker executing a **real** op def to completion and asserts `op.terminal` is published through the bus — i.e. the exact chain a live "run an op" check would exercise, minus the HTTP/SSE serialization frame.
- The remaining paths (failed, canceled-running, abandoned, force-dropped, canceled-queued) are each covered by a dedicated test in `terminal_events_test.go`.
- The only layer not exercised by tests is `Server.handleOperationsSSE`, which this PR **does not touch** (`git diff` over `internal/server/` is empty) — it just marshals `ev.Payload` and writes the SSE frame, unchanged.

If a reviewer wants a belt-and-suspenders live check, running any op in a dev instance with `curl -N .../api/v1/operations/events` should now show an `event: op.terminal` frame on completion where before there was none.

PR #2006, opened 2026-07-18
fix(maintenance): thread reporter progress + error return through remux/transcode/external-ID backfill (C2/H7)

jdfalk:
## Summary

T09 from `docs/agent-tasks/error-correction-2026-07/TASKS.md`: closes C2 and H7.

- **C2**: `internal/plugins/maintenance/backfill.go`'s remux and transcode ops
  logged one line, called a void dep method (`RemuxMalformedM4BFiles`/
  `TranscodeMalformedM4BFiles`) that could run for hours, then unconditionally
  logged "complete" — a 6-hour ffmpeg walk showed exactly one op-log line and
  failures couldn't fail the op.
- **H7**: same shape for the external-ID backfill
  (`internal/server/external_id_backfill.go` → `internal/itunes/backfill.go`)
  — the domain error was demoted to a Warn log, the op completed
  unconditionally, and the whole-library pagination reported no progress.

### What changed

- `internal/remux/remux.go` + `transcode.go`: `RemuxMalformedFiles`/
  `TranscodeMalformedFiles` now take a `func(processed, total int, msg string)`
  progress callback and return `error`. A cheap pre-count pass (same
  candidate predicate as the work pass) gives an accurate "X/Y"; progress
  fires every 25 files. A non-nil error is returned only for fatal setup
  failures (RootDir not configured, ffmpeg missing) — per-file failures stay
  non-fatal since transcode is the designed fallback for files remux can't
  fix. Added an explicit 30m `ProgressTimeout` to the transcode
  `OperationDef` since a full re-encode can exceed the registry's 5m default
  gap between progress stamps (remux doesn't need it — it advances on every
  candidate file, well inside 5 minutes).
- `internal/itunes/backfill.go`: `BackfillExternalIDs`/
  `BackfillITunesTrackPIDs` take the same callback and now propagate
  `GetAllBooksCore` read errors and `BulkCreateExternalIDMappings` write
  errors instead of silently `break`-ing the pagination loop or
  logging-and-discarding — both of those previously let the op mark itself
  "done" despite skipping part of the library.
- Dependency chain updated end to end: `maintenance.ServerDeps` →
  `internal/server/server_maintenance_deps.go` →
  `malformed_m4b_wrappers.go` / `external_id_backfill.go`. The UOS plugin Run
  functions in `internal/plugins/maintenance/backfill.go` wire the callback to
  `reporter.UpdateProgress` and fail the op on a non-nil error.
- The `startBackfills()` goroutines in `internal/server/server_lifecycle.go`
  run these same ops directly at every boot, **outside** the UOS op
  registry (no reporter available there). Added a small `startupProgressLogger`
  helper so this path — which is the one that's actually live at every
  startup — also surfaces more than a single log line across a multi-hour
  run, and now logs the previously-swallowed error via `slog.Warn` instead of
  it being unreachable.

### Flags for the reviewer / coordinator

- **Coordination**: this PR edits `internal/server/server_maintenance_deps.go`,
  which the T05 brief also owns. Unavoidable — changing the `ServerDeps`
  interface signature forces changing `*Server`'s implementation of those
  three methods, and they live in that file. The edit here is confined to
  the three one-shot-op methods (`BackfillExternalIDs`,
  `RemuxMalformedM4BFiles`, `TranscodeMalformedM4BFiles`), a different region
  from T05's author/metadata/transcribe edits, so the rebase should
  auto-resolve — but please sequence it.
- **Scope caveat**: I could not find where the UOS ops
  `maintenance.malformed-m4b-remux`, `maintenance.malformed-m4b-transcode`,
  or `maintenance.external-id-backfill` actually get enqueued (no
  `EnqueueOp` call site for any of the three). It's possible that path is
  dormant and these ops only ever run via the direct `startBackfills()`
  goroutines — that's why the `startupProgressLogger` fix to the goroutine
  path matters as the actual live-in-prod win, not just the UOS plugin
  wiring.
- **Left out of scope**: `remux.go`'s (and `transcode.go`'s) done-flag write
  happens unconditionally even when the walk was canceled mid-run,
  contradicting its own doc comment ("a canceled run does not write the done
  flag"). Real bug, not part of C2/H7 — flagging for a follow-up.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/remux/... -race -count=1` — pass (new tests for
      RootDir-not-configured error, progress-callback cadence/accuracy)
- [x] `go test ./internal/itunes/... -race -count=1` — pass (new tests for
      progress callback, bulk-write error propagation, read error propagation)
- [x] `go test ./internal/plugins/maintenance/... -race -count=1` — pass (new
      `backfill_ops_test.go` proving the Run functions thread progress to
      `reporter.UpdateProgress` and fail the op on a non-nil dep error)
- [x] `gofmt -l` clean on all 14 touched Go files
- [~] `go test ./internal/server/... -race -count=1` — timed out locally
      under heavy contention from ~6 sibling agents concurrently running
      their own `-race ./internal/server/...` suites on this machine (a
      dumped test, `TestDeleteBackup_Success`, was stuck on Pebble fsync
      disk I/O, unrelated to any code touched here — it passes in <5s in
      isolation). No `internal/server` test references any of the changed
      symbols. Deferring to CI for a clean-environment full run.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01WdKeHNKm5K6oyop4bYNd65

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 <!-- file: CHANGELOG.md -->
 <!-- version: 3.176.0 -->
+<!-- version: 3.175.0 -->
 <!-- guid: 8c5a02ad-7cfe-4c6d-a4b7-3d5f92daabc1 -->
 <!-- last-edited: 2026-07-18 -->
 
@@ -112,6 +113,30 @@
   operator can see unreadable groups accumulating.
 - All three packages (`internal/plugins/dedup`, `internal/itunes/...`,
   `internal/scanner`) pass `go test ./... -race -count=1`.
+#### July 18, 2026 - T09: remux/transcode/external-ID-backfill reporter threading (C2/H7)
+
+- **fix(maintenance):** the malformed-M4B remux op, its transcode twin, and the
+  external-ID backfill op each logged one "starting" line, called into a void
+  dependency method that could run for hours, then unconditionally logged
+  "complete" — a multi-hour ffmpeg walk gave no progress and could never fail
+  the op, even on a fatal setup problem (no RootDir configured, ffmpeg
+  missing) or a real persistence error.
+- Reworked the dependency chain
+  (`internal/plugins/maintenance/backfill.go` → `ServerDeps` →
+  `internal/server/malformed_m4b_wrappers.go` / `external_id_backfill.go` →
+  `internal/remux/remux.go` + `transcode.go` / `internal/itunes/backfill.go`)
+  so each now takes a `func(processed, total int, msg string)` progress
+  callback and returns `error`. The plugin wires the callback to
+  `reporter.UpdateProgress` and fails the op on a non-nil error.
+- Remux/transcode report an accurate "X/Y" (a cheap pre-count walk over the
+  same candidate predicate) every 25 files; external-ID backfill reports
+  after every page of book pagination and every 10,000 iTunes-XML tracks.
+  Per-file remux/transcode failures stay non-fatal (expected — transcode is
+  the designed fallback for files remux can't fix); a track-PID bulk-write
+  failure that was previously logged and discarded now fails the op.
+- No behavior change for the startup goroutines in `server_lifecycle.go`
+  that call these directly (outside the UOS op registry) — they now surface
+  the returned error via a Warn log instead of it being silently unreachable.
 
 #### July 17, 2026 - error-correction fix wave (15 PRs) + sandbox verification
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 <!-- file: CHANGELOG.md -->
 <!-- version: 3.176.0 -->
-<!-- version: 3.175.0 -->
 <!-- guid: 8c5a02ad-7cfe-4c6d-a4b7-3d5f92daabc1 -->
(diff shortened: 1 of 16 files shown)


jdfalk:
Update on the local \`internal/server\` test signal mentioned in the PR body: re-ran with an explicit 25m timeout to rule out a slow-but-real completion. Second confirmation it's pure contention, not a regression:

\`\`\`
panic: test timed out after 25m0s
\`\`\`

No \`--- FAIL:\`, no \`DATA RACE\`, no application panic — just the standard goroutine dump of idle worker-pool / Pebble-internal goroutines waiting on channels. Combined with the earlier finding (the one specific test caught mid-dump, \`TestDeleteBackup_Success\`, passes in <5s run standalone, and no \`internal/server\` test references any symbol this PR touches), I'm treating this as environmental (many sibling agents concurrently running the same package's \`-race\` suite on this machine) and deferring fully to CI.

No CI workflow had triggered for this branch as of this comment, despite CI clearly running for other concurrent PRs in this repo — likely queued behind the same burst of parallel-sweep activity. Will not merge regardless per the task instructions; leaving this for CI + reviewer to pick up.

PR #2007, opened 2026-07-18
fix(dedup): reroute book-merge off hard-delete to soft-delete path (F6/T10)

jdfalk:
### STEP 1 — VERIFICATION (recorded per brief)

**Question:** Does PebbleStore's `DeleteBook` tombstone external-ID (`ext_id:*`) mappings or enqueue iTunes ITL removals?

**Answer: NO — it does neither.** Traced `internal/database/pebble_store.go:2198 DeleteBook`. It deletes `book:*`, `book:path:*`, `book:versiongroup:*`, `book:work:*`, the three `book:hash|originalhash|organizedhash:*` rows, `metadata_state:*`, ISBN/ASIN index rows, and the `emb:v:book:*` embedding row. It **never** touches the `ext_id:*` keyspace (external-ID mappings live in `internal/database/pebble_store_externalids.go`, keys `ext_id:<source>:<id>` + `ext_id:book:<book>:<source>:<id>`) and **never** calls any write-back/EnqueueRemove path.

**Consequence of the legacy path** (`internal/dedup/book_dedup.go MergeBooks`, reached from `POST /audiobooks/merge` → `internal/server/handlers/duplicates/handler.go MergeBooks` → `internal/server/duplicates_ops.go` `RegisterBookMergeOp`): each merged-away loser is hard-deleted, so
- its `ext_id:<source>:<id>` primary mapping is left dangling, still resolving (via `GetBookByExternalID`) to a book row that no longer exists; a later iTunes/ASIN import resolving that PID lands on a deleted ID, and
- its iTunes tracks are never enqueued for ITL removal, so they live in the iTunes library forever.

The modern path (`internal/merge/service.go MergeBooks`) instead collects PIDs → `ReassignExternalIDs` (loser→winner) → `EnqueueRemove` (ITL) → `SoftDeleteBook` (recoverable). So reroute is required.

### STEP 2 — FIX: reroute (not documented-safe)

The `dedup.book-merge` op now routes through `merge.Service.MergeBooks` via a new extracted `applyBookMergeReroute` helper (`internal/server/duplicates_ops.go`). This reassigns external IDs to the winner, enqueues ITL removals, and soft-deletes losers — one merge path shared with the UI/version-group merge. Endpoint response shape is unchanged (the handler still returns an op-ID; only the call *inside* the op body changed).

**iTunes first-win copy preserved.** The 6 iTunes Book-level fields (persistent ID, play count, rating, date added, last played, bookmark) are NOT carried by `ReassignExternalIDs`; after soft-delete they would survive only on the loser row and vanish on a later purge/archive sweep. The wrapper copies them first-win onto the winner and persists BEFORE the merge (Service re-reads books fresh and writes the full keep object back, so the copy survives its version-group `UpdateBook`). The copy logic is extracted into a shared `dedup.TransferITunesMetadataFirstWin` helper so legacy and the reroute share one source of truth.

**Serialization preserved.** `itunes_heal` stays on legacy `dedup.MergeBooks` (via `merge.LockMergeRMW`), and the endpoint now uses `Service.MergeBooks` (`mergeSerializeMu.Lock`). Confirmed `LockMergeRMW()` locks the *same* `mergeSerializeMu` (`internal/merge/serialize.go`), so the two paths remain mutually exclusive on a shared row.

### STEP 3 — TEST

`internal/server/duplicates_ops_reroute_test.go` drives `applyBookMergeReroute` on a real `PebbleStore` and asserts: (1) loser is **soft-deleted** (`GetBookByID` non-nil, `MarkedForDeletion=true`) not hard-deleted; (2) `GetBookByExternalID` for the loser's PID now resolves to the **winner**; (3) winner inherited the loser's iTunes rating (first-win). Verified to **fail** against the legacy hard-delete behavior (loser goes nil, external ID left dangling).

### Behavior-change flags for the reviewer (non-blocking)

- **Hard-delete → version-group soft-delete.** Merged duplicates become soft-deleted version-group members (recoverable) instead of being deleted. Intentional per the T10 brief ("one merge path"), but a real product-behavior change for this endpoint.
- **Error semantics flip.** Legacy accumulated per-book errors and returned success (partial merge); `Service.MergeBooks` returns on the first bad ID, so one missing book now fails the whole op instead of merging the rest.
- **`itunes_heal` has the identical ext-ID/ITL gap** (hard-delete, no reassign, no ITL). Out of T10 scope — retained deliberately (it hard-collapses organize-bug rows) and flagged as a tracked follow-up in a code comment at the legacy call site.

### Tests
- `go build ./...` clean.
- `go test ./internal/dedup/... ./internal/merge/... -race -count=1` → all `ok`.
- `go test ./internal/server/ -race -run 'Merge|Dedup|Duplicat|Combine|Reroute|ExternalID' -count=1` → `ok 58s` (new test included, no races).
- Full `internal/server` package exceeds the 10-min default `go test` timeout under **and without** `-race` on this machine (documented pre-existing slowness; timeout lands inside unrelated E2E tests — my test and all tests up to it pass steadily). A long-timeout full `-race` run is in progress; CI runs the authoritative suite.

https://claude.ai/code/session_01WdKeHNKm5K6oyop4bYNd65

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 <!-- file: CHANGELOG.md -->
 <!-- version: 3.176.0 -->
+<!-- version: 3.175.0 -->
 <!-- guid: 8c5a02ad-7cfe-4c6d-a4b7-3d5f92daabc1 -->
 <!-- last-edited: 2026-07-18 -->
 
@@ -147,6 +148,24 @@
   callback, so this path also gets more than one log line across a
   multi-hour run, and now surfaces the previously-unreachable error via
   `slog.Warn`.
+#### July 18, 2026 - F6: reroute dedup book-merge off the hard-delete path
+
+- **Data-loss fix (T10 / F6):** the `POST /audiobooks/merge` endpoint
+  (`dedup.book-merge` op) was hard-deleting merged-away books via
+  `store.DeleteBook`, which does **not** tombstone external-ID (`ext_id:*`)
+  mappings and does **not** enqueue iTunes ITL removals. Every merge therefore
+  orphaned the losers' PID/ASIN lookups (leaving them resolving to a deleted
+  book) and stranded their iTunes tracks in the library forever. The op now
+  routes through `merge.Service.MergeBooks`, which reassigns external IDs to the
+  winner, enqueues ITL removals, and **soft-deletes** losers (recoverable) — the
+  same single merge path the UI/version-group merge uses. The legacy first-win
+  iTunes-stat copy (rating/play-count/etc.) is preserved onto the winner so the
+  reroute is strictly-better than the old path on every axis. The legacy
+  `dedup.MergeBooks` function is retained only for
+  `internal/reconcile/itunes_heal.go` (which intentionally hard-collapses
+  organize-bug duplicate rows and still carries the same ext-ID/ITL gap — tracked
+  as a follow-up). Regression test drives the reroute on a real PebbleStore and
+  asserts soft-delete + external-ID reassignment + first-win copy.
 
 #### July 17, 2026 - error-correction fix wave (15 PRs) + sandbox verification
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 <!-- file: CHANGELOG.md -->
 <!-- version: 3.176.0 -->
-<!-- version: 3.175.0 -->
 <!-- guid: 8c5a02ad-7cfe-4c6d-a4b7-3d5f92daabc1 -->
 <!-- last-edited: 2026-07-18 -->
 
--- a/TODO.md
+++ b/TODO.md
@@ -1,6 +1,5 @@
 <!-- file: TODO.md -->
 <!-- version: 10.4.1 -->
-<!-- version: 10.4.0 -->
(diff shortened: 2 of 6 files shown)


jdfalk:
### Update (post-review)

- **Integrity guard added.** `applyBookMergeReroute` now builds a deduped loser set that **excludes keepID** before both the iTunes transfer and the `merge.Service.MergeBooks` call. The handler binds `keep_id`/`merge_ids` without validation, so the keep book could appear in `mergeIDs`; passing it through would make the version-group loop write keep twice and demote it, leaving the group with no primary and keep neither primary nor soft-deleted. Legacy `dedup.MergeBooks` had a `keepID` skip; this restores that guarantee. New test `TestApplyBookMergeReroute_KeepIDInMergeIDs` covers it (keep survives as primary, loser soft-deleted).
- **Executive summary added** (data-loss fix): `docs/executive-summaries/2026-07-18-duplicate-merge-orphaned-itunes-links-executive-summary.md`.
- **Full `internal/server` `-race` run now confirmed green:** `ok ... 1025.317s` (exit 0, no data races). The earlier "timeouts" were purely the 10-min default `go test` timeout — the suite legitimately runs ~17 min under `-race` on this machine (pre-existing slowness, unrelated to this change).

PR #2042, opened 2026-07-24
feat(itunes): P0 verification — cross-type disjointness + field/bookmark preservation byte-proof

jdfalk:
## What

Adds `ComputeCrossTypeCollisions` (`internal/itunes/cross_type.go`) and a `--cross-type` mode to `cmd/pid-census`. This is the **disjointness backstop** for the 2-way-sync relocate op (P0 remaining measurement): an AO book_file PID must resolve to an **audiobook** track, never a music/podcast one — otherwise a relocate rewrites a non-audiobook track's location in the full-library `.itl`.

It classifies every `.itl` track (`isAudiobookITL`: Kind/Genre/Location) and cross-tabs against AO book_file ownership. Read-only; bounded worker pool for owner resolution.

## Result (measured on prod, 2026-07-24)

```
tracks_in_itl=97999  audiobook=92807  non_audiobook=5192
  ab_owned=81099  ab_unowned=11708  non_ab_owned=3436  non_ab_unowned=1756
CROSS_TYPE_COLLISIONS=3436  (all live-primary owner)
```

**The disjointness invariant HOLDS — 0 real cross-type collisions.** All 3,436 flagged tracks are audiobooks that `isAudiobookITL` under-classifies, not music. Proof: AO's DB stores only audiobooks (any AO-owned track is definitionally an audiobook), and the genre histogram over all 3,436 is 100% book-shaped — `Audio Book` (653), `(none)` (1130), `The First Law Book Two` (230), `Science Fiction`/`Sci Fi`/… (~660), `Suspense`, `Fantasy`, `Comedy`, `Speech`, `LGBT …` — **zero** music genres; kinds are ordinary mp3/aac/alac. The 1,756 `non_ab_unowned` are the user's real non-audiobook tracks, correctly hands-off.

→ The relocate op targets by AO book_file PID (already correct) and AO owns no music/podcast track, so a relocate can never make a cross-type write. **Backstop passes; no blocker for arming relocate (P2) on this ground.**

## Secondary finding (fail-safe, not fixed here)

`isAudiobookITL` misses `Audio Book`/`audio book` (it checks `"audiobook"` with no space — 705 tracks) and every literary-genre audiobook. For `GuardRebuildTarget` this is **fail-safe** (over-counts non-audiobooks → guard more likely to block, never less). But it must not be used as a relocate/cleanup targeting filter, and "fixing" it would *lower* the non-audiobook count and could weaken the rebuild guard's threshold — so any such change must re-derive those thresholds in the same PR. Tracked in `todo.d/`.

## Verification

- `go build ./...`, `go test ./internal/itunes/...`, `go vet` — clean.
- New `TestComputeCrossTypeCollisions` covers the 2×2 cross-tab incl. the non-audiobook-owned collision cell.
- Server artifacts (snapshot, work copy, staged binary) destroyed; fallback snapshot intact.

Findings: `docs/specs/2026-07-23-itunes-2way-p0-findings.md` §F5.

https://claude.ai/code/session_01J29y3VpN7FTczJmLeUJimt

### DIFF
--- a/changelog.d/itunes-2way-p0-cross-type-census.md
+++ b/changelog.d/itunes-2way-p0-cross-type-census.md
@@ -1,0 +1,22 @@
+<!-- file: changelog.d/itunes-2way-p0-cross-type-census.md -->
+<!-- version: 1.0.0 -->
+<!-- guid: 2f8b6c04-7d51-4e93-a1c6-9b3e5a0d1f72 -->
+<!-- last-edited: 2026-07-24 -->
+
+### Added
+
+#### iTunes cross-type PID-collision census — relocate disjointness backstop (read-only)
+
+`internal/itunes/cross_type.go` adds `ComputeCrossTypeCollisions`, and `cmd/pid-census` a
+`--cross-type` mode. It classifies every track in the AO writeback `.itl`
+(`isAudiobookITL`) and cross-tabs it against AO book_file ownership, surfacing the
+load-bearing invariant for the relocate op: an AO book_file PID must resolve to an
+audiobook track, never a music/podcast one. Bounded worker pool for owner resolution.
+
+Measured on prod (97,999 tracks): the disjointness invariant **holds** — 0 real cross-type
+collisions. The 3,436 tracks flagged are audiobooks that `isAudiobookITL` under-classifies
+(genre histogram 100% book-shaped, zero music genres); AO's DB stores only audiobooks, so no
+music/podcast track is AO-owned and a relocate can never make a cross-type write. Secondary:
+`isAudiobookITL` misses `Audio Book`/`audio book` (space) and literary-genre audiobooks —
+fail-safe for `GuardRebuildTarget` but not usable as a targeting filter. See
+`docs/specs/2026-07-23-itunes-2way-p0-findings.md` §F5.
--- a/changelog.d/itunes-2way-p0-preserve-byteproof.md
+++ b/changelog.d/itunes-2way-p0-preserve-byteproof.md
@@ -13,10 +13,17 @@ raw-byte comparison of the decompressed ITL payload before vs after a real reloc
 change to any atom the parser ignores — including the audiobook resume bookmark, which the
 binary parser does not parse. Env-gated (`ITL_PRESERVE_PROOF_PATH`); skips in CI.
 
-Proven on the real 97,999-track library: a relocate of 300 tracks changed **only** the
-location pair (0x0D/0x0B) — the full mith header (bookmark position, play count, rating,
-dates) and every other atom (Comment, Sort Name/Artist, content advisory, audio-data blob)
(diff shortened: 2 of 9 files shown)


jdfalk:
## Added: F6 — field/bookmark-preservation byte-proof

This PR now also carries the second remaining P0 verification (the branch is `feat/itunes-2way-p0-verify`).

`internal/itunes/itl_preserve_proof_test.go` (`TestITLPreservationByteProof`, env-gated `ITL_PRESERVE_PROOF_PATH`, skips in CI) does a per-track **raw-byte** comparison of the decompressed ITL payload before vs after a real relocate + remove. Raw bytes catch any change to any atom the parser ignores — including the audiobook resume **bookmark**, which the binary LE parser does not parse.

**Proven on the real 97,999-track library:**
```
relocated=300 (all non-location atoms byte-identical), removed=30, untouched-identical=97669
   (300 + 30 + 97669 == 97999 — exact partition, zero collateral mutation)
```
Per relocated track: mith header byte-identical except the 4-byte totalLen (the header holds the bookmark position, play count, rating, dates), and every mhod atom except the location pair (0x0D/0x0B) byte-identical — Comment, Sort Name/Artist, content advisory, audio-data blob. Remove touched only the 30 targeted tracks.

→ **Design §INV-F2 preservation is proven, not assumed.** Together with F5 (disjointness), the two safety pre-conditions for arming the relocate op (P2) now both hold. Findings: `docs/specs/2026-07-23-itunes-2way-p0-findings.md` §F6.

jdfalk:
## ⚠️ F7 — P2 blocker surfaced (owner-relevant)

Running the preservation proof through the **full contract path** (`UpdateITLLocations → SafeWriteITL`) uncovered a hard blocker for the P2 relocate op:

**The `location-form` safety guard rejects the entire live AO library — 82,976 tracks.** Every track's 0x0B URL contains `.itunes-writeback/` (e.g. `file://localhost/W:/audiobook-organizer/.itunes-writeback/iTunes Media/Audiobooks/…`), and the guard treats that substring as a staging-dir leak (the "damaged-4" class).

**These are not leaks — they're the AO library's real media paths.** The AO library physically lives at `W:\audiobook-organizer\.itunes-writeback\`, so iTunes' media folder is `…\.itunes-writeback\iTunes Media\`. In the locked hard-cutover design (iTunes pointed **at** the AO library), that substring is correct and unavoidable. `Force` does not override location-form (only the bounded-delta guard).

**→ The relocate op cannot write the live library until this is reconciled.** Preferred fix: scope the staging-marker check to the write *target* using the P0 4-state `LibrarySet` mode facts — reject `.itunes-writeback/` only when writing the **Original** library (or when the path's `.itunes-writeback/` root differs from the AO library's own root). Details + the reproduction test (`TestITLRelocateContractStatus`) in findings §F7 and a `todo.d/` fragment.

This is a P0-verification finding (a real blocker caught before P2 build), not a regression in this PR.

PR #2192, opened 2026-08-08
fix(operations): guard writeSetDeferred with r.mu — real data race in the dispatcher

jdfalk:
Found while diagnosing why the **Coverage Floor gate failed on #2189, a docs-only PR**. It wasn't a flake — it's a real data race introduced with last night's Gate 3b write-set conflict gate.

## The race

`-race` caught a concurrent map read at the prune in `dispatchCycle` against a `mapdelete` from a *second* `dispatchCycle` on the same `Registry`:

```
WARNING: DATA RACE
Read at 0x00c0029f14d0 by goroutine 14392:
  registry.(*Registry).dispatchCycle()  dispatcher.go:47
Previous write at 0x00c0029f14d0 by goroutine 14408:
  runtime.mapdelete_faststr()
  registry.(*Registry).dispatchCycle()  dispatcher.go:192
```

`Registry.writeSetDeferred` was read, ranged, and deleted from at **six sites** with no locking — even though `Registry` already carries an `mu sync.RWMutex` guarding its sibling state (`pluginRunning`, `concurrencyKeys`, `running`).

## Why the existing safety argument was wrong

The code asserted its own safety in a comment:

```go
// writeSetDeferred is only ever touched from the dispatcher goroutine
// (dispatchCycle is single-caller), so it needs no locking
```

That holds *within a single* `Start()`, which spawns exactly one dispatcher goroutine. But `Start()` is **deliberately restartable after `Shutdown()`** — it clears `notifyStopped` for precisely that case — and it neither waits for nor excludes a prior dispatcher goroutine that is still draining. During a restart, two `dispatchCycle` calls overlap on the same `Registry` and the assumption breaks. The comment has been replaced with one that explains this.

## The fix

All six accesses now take `r.mu`. Two details worth reviewing:

- **`logWriteSetDeferral`'s check-and-set uses one acquisition, not two.** Splitting it would let two goroutines both miss the entry and emit the same line twice — the exact log spam the dedupe exists to suppress (the 100ms ticker would otherwise repeat it ten times a second).
- **No lock ordering changed.** No call site already held `r.mu` — the Gate 3b path `RUnlock`s before calling the logger — so there's no deadlock risk from the added acquisitions.

## Verification

```
go test -race -run TestServerStartGracefulShutdown -count=3 ./internal/server/
  ok  github.com/falkcorp/audiobook-organizer/internal/server  44.734s

go test -race -short ./internal/operations/registry/
  ok  github.com/falkcorp/audiobook-organizer/internal/operations/registry  47.563s
```

Zero `DATA RACE` reports in either. `go build` and `go vet` clean.

No dedicated regression test is added — the race is already caught by the existing `-race` suite, which is what surfaced it. A deterministic reproducer would require forcing the restart-overlap window.

## Follow-up worth considering (not fixed here)

The deeper question is whether `Start()` should refuse to run, or wait on `goroutineWG`, while a prior dispatcher is still live. Two concurrent dispatchers on one Registry could double-dispatch operations in production, which the mutex makes memory-safe but not logically correct. Flagging rather than changing startup semantics at 3am.

### DIFF
--- a/changelog.d/20260808_024500_fix_dispatcher_writeset_race.md
+++ b/changelog.d/20260808_024500_fix_dispatcher_writeset_race.md
@@ -1,0 +1,36 @@
+### Fixed
+
+#### Data race on the dispatcher's write-set deferral map
+
+`Registry.writeSetDeferred` — the log-dedupe map added with the Gate 3b
+write-set conflict gate — was read, ranged, and deleted from at six sites in
+`internal/operations/registry/dispatcher.go` with no locking at all, while the
+`Registry` already carried an `mu sync.RWMutex` used for its sibling state
+(`pluginRunning`, `concurrencyKeys`, `running`). `-race` caught it as a
+concurrent map read at the prune in `dispatchCycle` against a `mapdelete` from
+a second `dispatchCycle` on the same `Registry`, failing
+`TestServerStartGracefulShutdown` and with it the Coverage Floor gate on an
+unrelated docs-only PR.
+
+The code carried an explicit comment asserting the map was safe:
+
+    // writeSetDeferred is only ever touched from the dispatcher goroutine
+    // (dispatchCycle is single-caller), so it needs no locking
+
+That is true within a single `Start()`, which spawns exactly one dispatcher
+goroutine — but `Start()` is deliberately restartable after `Shutdown()` (it
+clears `notifyStopped` for precisely that case) and neither waits for nor
+excludes a prior dispatcher still draining. During a restart two
+`dispatchCycle` calls therefore overlap on the same `Registry`, and the
+assumption breaks.
+
+All six accesses are now guarded by `r.mu`. The dedupe check-and-set in
+`logWriteSetDeferral` takes a single acquisition rather than two: splitting it
+would let two goroutines both miss the entry and emit the same line twice,
+which is the exact log spam the dedupe exists to suppress. No call site held
+`r.mu` already — the Gate 3b path releases it before logging — so no lock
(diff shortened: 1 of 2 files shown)


jdfalk:
Follow-up: I found the **exact mechanism** by which two `dispatchCycle` calls overlap, which makes the fix's justification concrete rather than hypothetical.

`Shutdown()` does not reliably join the dispatcher goroutine. From `registry.go` (~line 1006), in the comment explaining why the deps-sweeper is joined *explicitly*:

> the generic `goroutineWG.Wait()` below abandons stragglers after its **2s escape**

So `Shutdown()` can return with the dispatcher goroutine still live. `Start()` — which is explicitly written to support being called again after `Shutdown()` — then spawns a second dispatcher. Both run `dispatchCycle` against the same `Registry` and the same `writeSetDeferred` map.

That is not a test-only artifact: the 2s escape is deliberate, documented behaviour intended for stuck op goroutines. The old `// dispatchCycle is single-caller, so it needs no locking` comment was wrong for precisely this reason.

Production exposure is narrower — `Start()` is called once per process via `serviceregistry` (`container.go:197`), so a real double-start needs an in-process restart. The test harness does exactly that, which is why `TestServerStartGracefulShutdown` is where it surfaced.

This doesn't change the patch; the mutex is correct either way. It does reinforce the follow-up I flagged in the description: whether `Start()` should refuse to run, or join `goroutineWG`, while a prior dispatcher is still live. Two concurrent dispatchers are now memory-safe but could still double-dispatch operations.

PR #2282, opened 2026-08-10
fix(web): stuck closing menu left an invisible overlay swallowing all clicks [skip changelog]

jdfalk:
## What

Selecting an option from a MUI Select could leave the menu stuck in its `exiting`
state. MUI's Modal root is `position: fixed; inset: 0` and only becomes
`visibility: hidden` once the modal has **fully** exited — so a stalled exit
leaves a full-viewport element with `pointer-events: auto` sitting over the
page, swallowing every subsequent click until reload. The menu has already
faded to `opacity: 0`, so nothing looks wrong.

## Evidence

Reproduction (48-core host, 20 copies of the library-browser `clears all
filters` test across 12 workers). The stall rate tracks exit duration, which is
what makes this a **race**, not a dead timer:

| exit duration | passed / 20 |
|---|---|
| `'auto'` (~280ms) | **0** |
| 250ms | 8 |
| `exit: 0` | **20** |

Instrumented timeline: the exit *starts* correctly (paper reaches `opacity: 0`
at ~270ms) but the paper never receives the inline `visibility: hidden` that
Grow applies in the `exited` state — so react-transition-group sits in
`exiting` indefinitely, unchanged 4.8s later and through a full 30s timeout.

Isolated, the test passes 3/3. Under contention it failed 20/20. That contrast
is the negative control this fix is verified against.

## Not claimed

**Why RTG's completion callback never runs is unexplained.** Supplying a
numeric duration gives RTG its own fallback `setTimeout` and the stall still
happened 12/20 — which **disproves** the "MUI's auto-timeout is lost" theory I
started with. This change removes the window the race needs; it does not claim
to have fixed the race itself.

## Scope

- `MuiMenu` only, not `MuiPopover` — Autocompletes and pickers are untouched by
  a defect only observed on Select menus.
- `enter: 225` keeps the opening animation, which was never implicated.

## Also corrected

`playwright.config.ts` blamed "two workers on two pinned cores" and concluded
"neither the app nor the tests are wrong". Both halves are false: it reproduces
at **workers=1 on an idle 48-core host**, ~1 occurrence per 282 tests. That
comment is why this went unexamined twice. `workers: 1` on CI stays, on its own
merits.

## Counts

- Contention harness, `MuiMenu` fix: **20 passed / 0 failed**
- Same harness before the fix: **0 passed / 20 failed**
- Full chromium suite at gate config (`workers: 1`): running, will post

### DIFF
--- a/changelog.d/20260810_113000_fix_stuck_menu_blocks_page.md
+++ b/changelog.d/20260810_113000_fix_stuck_menu_blocks_page.md
@@ -1,0 +1,30 @@
+### Fixed
+
+#### A closing dropdown could leave the page unclickable until reload
+
+Choosing an option from a dropdown — the library filter selects are the ones
+this was found on — could leave the menu stuck part-way through closing. When
+that happened, every subsequent click anywhere on the page did nothing, and the
+only way out was to reload.
+
+Nothing looked wrong when it happened, which is what made it hard to spot. The
+menu had already faded to fully transparent, so the screen looked normal. But
+MUI's modal layer is a full-viewport element that only stops capturing clicks
+once the close has *completely* finished, and in this case it never did. So an
+invisible sheet stayed over the whole page and swallowed everything.
+
+The fix closes menus immediately instead of animating them out. Opening is
+unchanged — only the closing animation was involved, and only the closing
+animation is removed.
+
+For the record, since this was mis-diagnosed twice before: the failure is a
+race, not a broken timer. Measured on a 48-core host, running 20 copies of the
+affected end-to-end test across 12 workers, the failure rate tracked the length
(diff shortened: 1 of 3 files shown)


jdfalk:
Follow-up, for the record: the full chromium suite at gate config (`workers: 1`) on this commit came back **277 passed / 1 failed / 4 skipped**. The failure *moved* rather than disappearing — from the Select menu to `.MuiDrawer-modal .MuiBackdrop-root`, i.e. the same stuck-modal defect on a different component. Fixed in #2283, after which the suite is **278 passed / 0 failed / 4 skipped**.

That also refuted this PR's stated mechanism: Drawer already used numeric durations (so it had react-transition-group's fallback timer all along) and stalled anyway at 17/20. See #2283.

PR #2291, opened 2026-08-11
feat(web): MUI 5.14 → 6.5 (upgrade step 1) [skip changelog]

jdfalk:
MUI `5.14` → `6.5.0`. **Draft — one confirmed regression blocks it.**

## Correction to an earlier version of this PR body

This body previously stated that `main` was red at 555 passed / 1 failed / 8 skipped. **That was wrong and is retracted.** It rested on a single run. A re-run of the same SHA (`76269d57`) returned **556 passed / 0 failed / 8 skipped, exit 0**, with `batch-operations.spec.ts:100` [webkit] passing in 1.6s. **`main` is green.** That spec is an intermittent webkit flake, not a standing failure.

## Attribution, now n=2 per side

| Spec | main | this branch | Verdict |
|---|---|---|---|
| `scan-import-organize.spec.ts:259` [chromium] | ✅ 2/2 | ❌ **2/2** | **real v6 regression** |
| `batch-operations.spec.ts:100` [webkit] | ✅ 1, ❌ 1 | ✅ 1, ❌ 1 | flake, unrelated to v6 |

Same machine, quiet, full `make test-e2e` each time. Branch run 2: 555 passed / 1 failed / 8 skipped, exit 2.

## What is done and green

- `@mui/material` + `@mui/icons-material` at 6.5.0
- `v6.0.0/sx-prop` codemod — 40 files, 0 errors; `v6.0.0/theme-v6` needed no changes
- `npm run build` clean; **vitest 65 files / 448 tests pass**
- Emotion stays the engine; Pigment CSS deliberately not adopted

The predicted v6 ripple/`act()` churn never materialised, because the 2026-08-07 inventory holds: 0 `styled()`, 0 JSS, 0 theme augmentation, 0 deep imports, no `@mui/lab`, no `@mui/x-*`. All 1,726 `sx=` usages carry over.

## The blocker

`scan-import-organize.spec.ts:259` [chromium]:

```
expect(locator).toHaveCount(expected) failed
locator('.MuiDrawer-modal .MuiBackdrop-root').filter({ visible: true })
Expected: 0   Received: 1   Timeout: 15000ms
  - 34 × locator resolved to 1 element
```

That assertion is the tripwire added in #2283 after a stuck Drawer left an **invisible** full-viewport backdrop swallowing every click until reload. Under v6 the backdrop stays visible the full 15s — the close transition never completes.

**It will not be relaxed or deleted to get v6 green.** That is precisely the move the incident exists to prevent, and `toBeHidden()` cannot see an `opacity:0` element that still eats clicks.

Passes in isolation (32/32), fails only in the full suite — the same profile as the original incident, whose root cause was never explained and is labelled so in-code. v6 reworked Modal/Backdrop transitions; the candidates to investigate are the backdrop's `transitionDuration` default, `keepMounted`, and `closeAfterTransition`.

### DIFF
--- a/changelog.d/20260810_205000_mui_v6_upgrade.md
+++ b/changelog.d/20260810_205000_mui_v6_upgrade.md
@@ -1,0 +1,22 @@
+### Changed
+
+#### Material UI upgraded from 5.14 to 6.5 (step 1 of the path to v9)
+
+This is the first of three version hops that bring the interface library up to
+date. Nothing about the app should look or behave differently — this step exists
+to move the foundation, not to change the UI.
+
+Two automated migrations were applied: one that rewrites conditional styling
+into the form version 6 expects (40 files), and one for the theme file, which
+turned out to need no changes at all. Nothing was migrated by hand.
+
+The upgrade was unusually quiet, and the reason is worth recording: an inventory
+taken on 2026-08-07 found this codebase uses none of the patterns that normally
+make these upgrades painful — no legacy styling engine, no custom theme type
+extensions, no deep imports into library internals, and none of the optional
+add-on packages that have to move in lockstep. The 1,726 inline style props
+throughout the app carry over untouched.
+
+Notes for the next step: the real cost in this migration is the grid layout
+rewrite, which lands in the 6→7 hop, not this one. Emotion remains the styling
+engine — the zero-runtime alternative is deliberately not being adopted.
--- a/changelog.d/20260811_080000_fix_drawer_stuck_backdrop_exit_false.md
+++ b/changelog.d/20260811_080000_fix_drawer_stuck_backdrop_exit_false.md
@@ -1,0 +1,32 @@
+### Fixed
+
+- **The filter Drawer could stall mid-close and leave an invisible full-viewport
+  backdrop that swallowed every click until reload — and the 2026-08-10 "fix"
+  for it had only narrowed the race, not closed it.** `web/src/theme.ts` now
+  sets `MuiDrawer.defaultProps.slotProps.transition = { exit: false }`, which
+  makes react-transition-group take its synchronous branch in `performExit`
+  (`safeSetState({status: EXITED})` from inside `componentDidUpdate`) instead of
+  deferring completion to a `setTimeout`. Instrumented traces show the deferred
+  update is the thing that gets lost: the timer fires, RTG calls `setState` on a
+  mounted instance, and React never applies it — the same instance re-renders
+  4 ms later still reporting `exiting`, and is still reporting `exiting` 300 ms
+  later. `onExited` therefore never runs, MUI's `Modal` never unmounts, and its
+  `position: fixed; inset: 0` backdrop keeps eating clicks while looking
+  perfectly closed.
+
+  Measured on chromium at `workers=1`, n=10 per cell. Before, on the MUI 6.5.0
+  branch: an Escape with no preceding CDP round-trip stalled **10/10**. After:
+  **0/10**, and the real gate (`scan-import-organize.spec.ts` "complete
+  workflow") went from failing to **10/10 passing**.
+
+  The previously shipped `transitionDuration: { exit: 0 }` is retained for the
+  enter animation but is *not* the fix; on `main` (MUI 5.18.0), which already
+  carried it, the same defect still reproduced 9/10 when the Escape was preceded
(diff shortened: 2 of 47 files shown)


jdfalk:
## v6 blocker — narrowed, not yet solved (2026-08-11 ~02:40)

Reproduced at n=3, workers=1, on an otherwise idle host, so this is not contention:

```
✘ scan-import-organize.spec.ts:259 [chromium] complete workflow   (2 of 3 shown failing)
  expect(locator).toHaveCount(expected) failed
  Locator:  locator('.MuiDrawer-modal .MuiBackdrop-root').filter({ visible: true })
  Expected: 0   Received: 1
```

### Three theories tested and DISPROVEN

**1. "v6 stopped forwarding `transitionDuration` to the backdrop."** False. Read the installed v6 source: `Drawer.js:236` builds the backdrop slot with `mergeSlotProps(slotProps.backdrop || {...}, { transitionDuration })`, and `@mui/material/utils/mergeSlotProps` merges `{...defaultSlotProps, ...externalSlotProps}` — so with no external backdrop props the theme's `{enter:225, exit:0}` **does** reach `Backdrop`, which passes it to `Fade` as `timeout` (`Backdrop.js:103`). The plumbing is intact.

**2. "Escape no longer closes the Drawer under v6."** False. `library-browser.spec.ts` › `clears all filters` opens the same `FilterSidebar` drawer, interacts with the same Select, presses Escape — **3/3 pass** on this branch.

**3. My first discriminating probe.** Inconclusive and discarded: it timed out on `getByRole('button', {name:/filters/i})` because I never seeded the library, so *both* arms failed for a reason unrelated to Escape. Recording it because the run looked like strong evidence for theory 2 and was not.

### What IS established

- The failure is **specific to `scan-import-organize`'s state**, not to Drawer/Escape/transitions generally.
- At failure the drawer content is still in the DOM (`Library State` present in the captured a11y snapshot) and one `.MuiDrawer-modal` backdrop is visible.
- `Filters 1` is marked `[active]` in that snapshot — a filter is applied, which the passing spec's path does not necessarily reproduce.

### Next step for whoever picks this up

Diff the two paths. `library-browser` reaches the drawer on a plain `/library` load; `scan-import-organize` reaches it after a full settings→scan→navigate flow with a mocked EventSource. The candidate difference is a second modal (the scan/organize dialog) whose unmount interacts with the drawer's, or a re-render triggered by the mocked operation stream while the drawer is closing.

**Do NOT relax the assertion.** It is the tripwire from #2283 and it is currently telling the truth: a full-viewport backdrop is left visible and swallowing clicks, which is a dead page for the user.

jdfalk:
## Ownership-verified before/after on the real gate

The earlier numbers in this PR were measured without proving which worktree owned `:8484` — eleven worktrees on this machine all default to that port, so a sibling's server can serve another's run. Re-measured with a harness that kills the port, starts the run, and records the listener's PID, command line and cwd before any assertion executes.

**Spec:** `scan-import-organize.spec.ts` → `complete workflow: add import path → scan → organize`, chromium, `--workers=1 --repeat-each=10`.

| arm | result | listener cwd (proven) | playwright exit |
|---|---|---|---|
| v6.5 + `exit: 0` only (control) | **4 passed / 6 failed** | `…/audiobook-organizer-mui6/web` | 1 |
| v6.5 + `exit: false` (this PR) | **10 passed / 0 failed** | `…/audiobook-organizer-mui6/web` | 0 |

n=10 per arm. Both proofs are in the run artifacts; each names the listener PID and confirms its cwd is this worktree, not a sibling.

## Why this is structural and not a retune

The discriminating cell is the **v5** arm. `exit: 0` was demonstrably leaky on v5.18 too (P1 9/10, P2 1/10 — the rate *moves with the probe*, which is the signature of a race window rather than a weak fix). `exit: false` closes it on **both** versions, 10/10 in all four cells.

That also corrects the framing this PR started with: **v6 did not introduce the defect.** It moved the timing window onto the common path. The original comparison that read as a v6 regression was n=2.

## What is not claimed

Why React silently drops a `setState` issued from a `setTimeout` on a mounted class component inside a portal is **still unexplained**, and is labelled as unexplained in `web/src/theme.ts` rather than papered over. Duplicate React copies, StrictMode, `flushSync`/`startTransition`, uncaught exceptions and unmounting are all ruled out.

Rebased onto `main` at `bcc4f0fa`.

PR #2322, opened 2026-08-12
ci: raise the Go Tests cap upstream, bump the pin, and correct my wrong attribution [skip changelog]

jdfalk:
I merged #2317 tonight claiming it fixed the repeated `conclusion=cancelled` failures on this repo's Go tests. It did not, and the comment making that claim is currently sitting on main where it will mislead whoever reads it next.

## What actually happens

The check that keeps getting cancelled is `Minimal CI / Go Tests (short, race)`. **That job is not defined in this repository.** The `Minimal CI /` prefix on a check name means the job comes from a called workflow — here `falkcorp/github-common/.github/workflows/reusable-ci-minimal.yml`, pinned at `d0c3326b`. In that file:

```
100:    name: Go Tests (short, race)
102:    timeout-minutes: 20
137:    run: go test -short -race -timeout 30m ./...
```

A 20-minute job cap against a 30-minute `go test` timeout. The runner kills the job ten minutes before Go's own timeout can fire, and Go's `-timeout` is the only thing that prints a goroutine dump naming the stuck test. So the run ends as bare `cancelled` with no evidence, and `cancelled` renders on the PR as a test failure.

## The measurement that settles it

From run `31561741823` on #2319 — a docs-only PR:

```
cancelled   Minimal CI / Go Tests (short, race)   03:58:53Z -> 04:19:09Z   (20m16s)
success     Coverage Floor (PR gate)              03:58:48Z -> 04:12:09Z   (13m21s)
```

`Coverage Floor (PR gate)` is the job #2317 raised to 35. It finished comfortably, in the very same run where the other job was killed at its cap. It has never been the job being cancelled.

## What this PR changes

Comment only — no behaviour. The `timeout-minutes: 35` stays, because it is independently correct for that job on its own terms (`make test-short` runs with `-timeout 25m`, so 35 > 25 satisfies the invariant). What changes is that the comment no longer claims credit for a bug in another repository, and it now records where the real cap lives so the next person doesn't repeat my grep.

The real fix is upstream in **falkcorp/github-common#346**. It could not have been made here: that workflow exposes no timeout input (`go-version`, `node-version`, `go-experiment`, `system-packages`, `frontend-working-dir`, `run-frontend`).

## Why I got it wrong

I grepped `timeout-minutes` in `ci.yml`, found one, and changed it. The grep answered "is there a timeout in this file" when the question was "what caps *that* job." The `/` in the check name was the available tell and I didn't read it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 # file: .github/workflows/ci.yml
-# version: 3.3.0
+# version: 3.3.1
 # guid: f1a2b3c4-d5e6-f7a8-b9c0-d1e2f3a4b5c6
-# last-edited: 2026-08-11
+# last-edited: 2026-08-12
 
 # Fast parallel CI for PRs and pushes.
 # Runs go vet/build, short tests, frontend lint/build, and frontend unit tests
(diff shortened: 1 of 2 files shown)


jdfalk:
Updated: this PR now also **bumps the reusable-workflow pin** to `85a2909` (the merge of falkcorp/github-common#346), so the 35-minute cap actually takes effect here. Without the bump this repo keeps calling the old `d0c3326b` and every slow run keeps ending as `cancelled` with no goroutine dump.

Verified by content at the merge SHA:

```
100:    name: Go Tests (short, race)
115:    timeout-minutes: 35
150:        run: go test -short -race -timeout 30m ./...
```

35 > 30, so Go's timeout can now fire and name the stuck test.

PR #2336, opened 2026-08-12
fix(database): serialise EmbeddingStore.Close against in-flight operations [skip changelog]

jdfalk:
Fixes the hang diagnosed in #2329. Closes out the "Minimal CI / Go Tests — cancelled, no reason given" runs.

## The bug

`pebble.DB.Close()` running while operations are in flight. Three writers already inside `db.Set` block on Pebble's commit-pipeline mutex and never wake:

```
goroutine 2261 [sync.WaitGroup.Wait, 29 minutes]   <- the test's wg.Wait()
3 x writer     [sync.Mutex.Lock,    29 minutes]
    pebble/v2.(*commitPipeline).prepare   commit.go:455
    pebble/v2.(*DB).Set                   db.go:646
    database.(*EmbeddingStore).setJSON
```

Pebble documents that calling `Close` concurrently with any other DB method is unsafe. This change makes `EmbeddingStore` provide the guarantee Pebble doesn't.

**Reached only by tests.** `NewEmbeddingStore` sets `owned: false`, so production's `Close()` returns on its first line and never touches Pebble. The cost was paid entirely in CI.

## Why the two existing defences could never work

- **`recover()`** is in all three chaos tests. A deadlock is not a panic — `recover()` cannot see it. That is why the tests looked adequately guarded for months.
- **`closed atomic.Bool`** is a check-then-act. A writer already inside `db.Set` is past the check. Only a lock held for the operation's **duration** closes that window.

## The change

Every method that touches `s.db` takes `closeMu.RLock` for its whole duration; `Close` takes the write lock, so it cannot begin until the last in-flight operation has returned. **21 methods guarded.** No `closed` check was added to them — Pebble already returns `ErrClosed` on a closed DB (which is why `TestChaos_OperationsAfterClose` passes today), so the lock only needs to provide exclusion, not error reporting. That kept the diff uniform instead of requiring 21 different zero-value returns.

Deliberate exceptions, derived from the internal call graph rather than assumed:

| method | treatment | why |
|---|---|---|
| `nextID` | unguarded | only caller `UpsertCandidateNew` already holds RLock |
| `listCandidatesByStatusIndex` | unguarded | only caller `ListCandidates` already holds RLock |
| `IsCandidateStatusIndexBuilt` | **split** into wrapper + `isCandidateStatusIndexBuiltLocked` | exported *and* called by `ListCandidates` under RLock |

That third one matters. `sync.RWMutex` does not permit recursive `RLock`: a writer blocking between the two acquisitions leaves the second waiting on a writer that is waiting on the first. Guarding it naively would have traded a rare deadlock for a **guaranteed** one — and would have passed every test that never calls `Close`.

`FindSimilar`, `UpsertCandidate` and `HealthStats` call other exported methods but touch no DB primitive themselves, so they correctly take no lock at all.

Lock ordering is consistent everywhere: `closeMu` first, then the pre-existing `mu`. Nothing takes them in the other order.

## Second layer: a hang now reports itself

The deadlock was one silent layer; **a hang producing no test failure at all** was the other. A bare `wg.Wait()` meant a regression showed up as a dead CI job with no test named and no stack. The three chaos tests now use a bounded `waitForWorkers`, so a regression fails in 30 seconds and says which invariant broke.

## Verification

**The guard was watched going red before being trusted green.** Injecting a worker that outlives the deadline:

```
--- FAIL: TestChaos_MixedReadWriteDuringClose (30.13s)
    workers still running 30s after Close returned: operations raced Close and
    deadlocked. EmbeddingStore.closeMu is supposed to make this impossible ...
```

Then removed (`grep -c NEGCTRL` = 0) and re-run clean.

- `go build` — exit 0
- `go vet` — exit 0
- `go test ./internal/database/ -run TestChaos -count=30 -race` — **exit 0, 66.0s**. This is the check that matters for the new lock: 30 consecutive runs with no recursive-`RLock` regression.
- Full `./internal/database/` suite with `-race` — running; will report the count before merge.

## What is NOT claimed

**The original hang was never reproduced locally — 0 of 50 runs on macOS with `-race`.** So this fix is justified *by construction* (the write lock makes concurrent `Close` impossible) and by no-regression evidence, **not** by observing the hang disappear. macOS scheduling does not produce the Linux + `-race` + parallel-package interleaving that CI does.

The rate remains unquantified. #2329's "4–5%" was retracted after the same branch hung 2 of 2 attempts; this PR also fixes the heading that still carried that number.

`PebbleDB()` still hands the raw handle to adjacent stores, which is outside this lock. Pre-existing, unchanged, and out of scope here.

### DIFF
--- a/changelog.d/20260812_fix_embeddingstore_close_deadlock.md
+++ b/changelog.d/20260812_fix_embeddingstore_close_deadlock.md
@@ -1,0 +1,25 @@
+### Fixed
+
+#### `EmbeddingStore.Close()` could deadlock against in-flight operations
+
+Closing an `EmbeddingStore` that owns its PebbleDB while operations were still running
+could hang forever: writers already inside `db.Set` blocked on Pebble's commit-pipeline
+mutex and never woke. Pebble documents that calling `Close` concurrently with any other DB
+method is unsafe; the store now provides the guarantee Pebble does not.
+
+Every operation that touches the database holds a read lock for its whole duration and
+`Close` takes the write lock, so `Close` cannot begin until the last in-flight operation
+has returned.
+
+This was reached only by tests — the production constructor `NewEmbeddingStore` sets
+`owned: false`, so `Close()` returns immediately and never shuts the DB down. Its cost was
+paid in CI, where `TestChaos_MixedReadWriteDuringClose` hung for the full job timeout and
+reported only an unexplained "cancelled".
+
+The pre-existing `closed` flag could not have prevented this and was never going to: it is
+a check-then-act, and an operation already inside `db.Set` is past the check.
+
+Also bounded the three chaos tests' worker waits. A deadlock is not a panic, so the
+`recover()` in each of them could never catch this, and a bare `wg.Wait()` turned a
+regression into a silent 30-minute hang with no test named. A regression now fails in 30
+seconds and says which invariant broke.
--- a/internal/database/embedding_store.go
+++ b/internal/database/embedding_store.go
@@ -1,6 +1,6 @@
 // file: internal/database/embedding_store.go
-// version: 2.9.0
-// last-edited: 2026-07-17
+// version: 2.10.0
+// last-edited: 2026-08-12
 // guid: 7c4a9b2e-d831-4f5c-a07e-3b8d6e1f9c42
 
 package database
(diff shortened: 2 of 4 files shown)


jdfalk:
## Further evidence: this is blocking two other PRs, in two different jobs

While waiting on CI here, #2321's `Coverage Floor (PR gate)` failure turned out to be **the same deadlock**, not an unrelated coverage problem:

```
FAIL	github.com/falkcorp/audiobook-organizer/internal/database	1500.036s
make: *** [Makefile:180: test-short] Error 1
```

1500.036s is exactly 25 minutes — `make test-short`'s timeout. Same package, same shape, different job. So the hang is not confined to `Minimal CI / Go Tests`; it fires anywhere `internal/database` runs.

**Four observed occurrences today, across three PRs:**

| PR | job | duration | outcome |
|---|---|---|---|
| #2323 | Go Tests | 33m | 30m Go timeout, named the test |
| #2323 (re-run) | Go Tests | 34m03s | 30m Go timeout again |
| #2321 | Coverage Floor | 25m00s | `make test-short` timeout |
| (earlier, run history) | Go Tests | 20m14s | old 20m cap |

That is a good deal more frequent than the "4–5%" figure I published and then retracted in #2329 — and it is a second, independent reason not to quote a rate: I had been counting only one job.

Consequence: merging this unblocks **#2321 and #2323 together**. Neither has any defect of its own; #2321's other red check is the `%T` CodeQL false positive already analysed, and #2323 is a docs-only change.

PR #2337, opened 2026-08-12
test(abs): conformance suite checked shape but never values

jdfalk:
## What

The 22 ABS golden-fixture assertions ran with `conformance.Options{IgnoreExtra: true}` and
**no `CompareValues`**. That compares key presence and type. A handler returning every
correct key with entirely wrong data passed every one of them.

`CompareValues: true` is now on by default. **Ten endpoints already met the oracle exactly**
and are pinned to it from this commit: ping, status, libraries, series, authors, narrators,
paginated authors, and the three bookmark routes.

## The part that isn't the one-line flip

Strictness was opt-in, so a newly added endpoint got the weak check **for free** and turning
it up required someone to remember. That is the same shape of defect as #2333's "grep for
`/api/v1` twins" rule: a correct instruction whose enforcement is a human habit.

Inverted. Strict is the default; weakening it costs effort and leaves evidence. The twelve
call sites that cannot meet the oracle yet must name themselves via `assertConformantPending`
**and supply a written reason** — an empty reason fails the test rather than silently buying
another release of weak checking:

```go
if strings.TrimSpace(reason) == "" {
    t.Fatalf("%s: assertConformantPending requires a reason — a call site that cannot say "+
        "why it is exempt has no business being exempt", fixture)
}
```

The remaining N-2 work is therefore enumerated **in the source**, not in a tracking doc, and
the count cannot drift upward quietly.

Note that `assertConformantPending` is shape-only for **values**. Shape findings still fail
it — the exemption is narrow on purpose, so a pending call site cannot also start returning
the wrong keys.

## The twelve are not one problem

Measured by flipping both helpers and running the whole package:

| Pile | Count | What it is |
|---|---|---|
| Fixture drift | 8 | The fake library's multi-file book is synthetic — six identical 1662 s tracks, 2049–2054 byte files, `timeBase 1/1000` — where the oracle captured a real *Odyssey* recording: six distinct durations summing to 9975.48, files of 11–21 MB, `timeBase 1/14112000` |
| Identity divergence | 4 | `user.type` root/user · `Source` docker/audiobook-organizer · `permissions.upload` and `.createEreader` want true, got false. Plausibly ours to keep |
| **Genuine mapper behaviour** | **2** | `metaTags.tagTrack` — we pass raw ID3 `TRCK` `4/24` through where ABS normalizes to `4` · audio-track `title` — we send the tag title, ABS sends the filename |

The last two are the reason this was worth doing. They are **live mapper behaviour on a
client-facing surface**, they are not test data, and shape-only checking could not see them:
both fields were present and both were strings.

## Verification

The gate was watched going **red**, not merely observed green — a green conformance run is
exactly what the old shape-only suite produced for two years.

Corrupting one seeded author name (`"Homer"` → `"Homerr"`) turns the strict authors
assertions red and names the field. Restored before commit.

`go vet ./internal/server/handlers/abs/` exit 0 · full package `ok 23.743s`, **no `-run`
filter** — the #2335 CI failure was a test my filter did not match.

## Not in this PR

- **Reseeding the fake library from the oracle** so drift becomes structurally impossible.
  That is the follow-up, and it is what retires most of the twelve.
- **The two mapper questions**, which need a call on real client behaviour rather than a
  reading of our code.
- **N-4's residual `/api/authors/:id` redirect.** Closing it means gating the ABS surface on
  `ABSAPIEnabled` with explicit 404 handlers inside the group — a routing change with
  production blast radius that has no business riding along with test-fixture work.

### DIFF
--- a/changelog.d/20260812-abs-conformance-values.md
+++ b/changelog.d/20260812-abs-conformance-values.md
@@ -1,0 +1,31 @@
+### Changed
+
+- **The ABS conformance suite now checks values, not just shape.** All 22 golden-fixture
+  assertions ran with `conformance.Options{IgnoreExtra: true}` and no `CompareValues`,
+  which compared key presence and type and nothing else — a handler that returned every
+  correct key with entirely wrong data passed. The suite read as a conformance gate and
+  gated almost nothing.
+
+  Ten endpoints already met the oracle exactly and are now pinned to it: ping, status,
+  libraries, series, authors, narrators, paginated authors, and the three bookmark
+  routes.
+
+- **Strictness is now the default, and weakening it is what costs effort.** The previous
+  arrangement had this backwards: a newly added endpoint got the weak check for free and
+  turning it up required someone to remember. The twelve call sites that cannot meet the
+  oracle yet must now name themselves via `assertConformantPending` **and give a written
+  reason** — an empty reason fails the test. Those twelve are the remaining work, they
+  are enumerated in the source rather than in a tracking doc, and the count cannot drift
+  upward quietly.
+
+  Their failures are not one problem. Eight are fixture drift: the fake library's
+  multi-file book is synthetic (six identical 1662 s tracks, 2049–2054 byte files,
+  `timeBase 1/1000`) where the oracle captured a real recording of *The Odyssey* (six
+  distinct durations summing to 9975.48, files of 11–21 MB, `timeBase 1/14112000`).
+  Four are identity divergences we may well intend to keep — `user.type`, `Source`,
+  `permissions.upload`, `permissions.createEreader`.
+
+  Two are neither, and would have stayed invisible under shape-only checking: we emit
+  the raw ID3 `TRCK` value `4/24` in `metaTags.tagTrack` where AudiobookShelf normalizes
+  it to `4`, and we send the tag title for an audio track where AudiobookShelf sends the
+  filename. Both are live mapper behaviour rather than test data.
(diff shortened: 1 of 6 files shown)


jdfalk:
## Correction to this PR's description

The description claims **two** genuine mapper divergences. Only **one** is real.

**`metaTags.tagTrack` is fixture drift, not a mapper defect.** I read the single diff row (`want 4, got 4/24`) and inferred a normalization rule. Reading all six oracle tracks shows there is no rule:

```
index=1 tagTrack='1/24'   index=4 tagTrack='4'
index=2 tagTrack='2/24'   index=5 tagTrack='5/24'
index=3 tagTrack='3/24'   index=6 tagTrack=None
```

ABS is passing the raw ID3 tag through — track 4's real tag was a bare `4`, track 6 had none. `mapper.go:758` lets `RawTags["track"]` win over the computed `n/total`, which is the same behaviour. The difference exists only because the seed hardcodes `fmt.Sprintf("%d/24", i)` for `i < 6`, manufacturing a value the real file never carried.

**The track-title finding stands.** `media.tracks[].title` is `odyssey_01_homer_butler_64kb.mp3` in the oracle while `metaTags.tagTitle` separately carries `The Odyssey: Book 01`. `trackTitle()` (`mapper.go:605`) prefers `BookFile.Title` and only falls back to the basename, so we send the tag title where ABS sends the filename.

Corrected in the changelog fragment and in the `get_api_items_id.json` pending reason on the follow-up branch, so the wrong claim does not reach `CHANGELOG.md` at release time.

The generalizable error: **one row of a diff is one sample.** The column had the answer in it.

PR #2339, opened 2026-08-12
fix(organize): report failures and honour ctx instead of always returning nil [skip changelog]

jdfalk:
Three defects on one path, all of the same shape: an organize run describing itself as something other than what happened.

## 1. It always reported success

`PerformOrganize` ended in an unconditional `return nil`. **Every book could fail and the caller still saw success**, so the operation was recorded as succeeded and nothing upstream had any way to learn otherwise.

The caller in `library_core_ops.go` already had a correct error path — `op.SetStatus("failed")`, log, return — it was simply **unreachable**.

## 2. The summary hid failures

The logged summary listed organized / re-organized / already-correct / skipped, but **not failed**. `stats.Failed` existed only in the `organize_summary` operation-change row, which nobody reads. So a run where every book failed printed:

```
Organize complete: 0 organized, 0 re-organized, 0 already correct (stamped), 0 skipped
```

Indistinguishable from a run that had nothing to do. A cancelled run also said "complete".

## 3. `ctx` was accepted and never read

`organizeBooks(ctx, ...)` never referenced `ctx` — its only occurrence was the parameter itself. Cancellation was checked **only in the job feeder**, so stopping a run stopped new work being queued but let the eight workers drain everything already buffered. Context cancellation — a disconnected HTTP client, or server shutdown — did nothing whatsoever.

## The rule now, stated deliberately

| outcome | result |
|---|---|
| cancelled | error wrapping `ErrOrganizeCanceled` (checked **first**) |
| every attempted book failed, none succeeded | error |
| partial failure (1 of 3000) | **success** — count carried by the summary |
| nothing to do | success |

Partial failure stays a success on purpose: one failure in three thousand books must not fail the whole operation. Cancellation is checked before total failure because a cancelled run has an *unknown* outcome, not a bad one — books it never reached are neither organized nor failed, and calling that "failed for all N books" would be its own misreport.

## The trap this avoids

Making cancellation return an error would have swapped one misreport for another: the caller marks any non-nil error `"failed"`, so a deliberate cancel would have been recorded as a failure. Hence `ErrOrganizeCanceled` and an `errors.Is` check before the failure branch. The v2 registry worker already gets this right (it checks `ctxCanceled` before `runErr`); this only aligns the logged op status with it.

## Verification

`organizeOutcomeError` and `formatOrganizeSummary` are extracted as pure functions so the policy and the user-visible text are **stated and testable** rather than implied by the last line of a 100-line method.

- `go build`, `go vet` on `internal/organizer` and `internal/server` — **exit 0**
- `go test ./internal/organizer/` — **exit 0**, 5 tests / 11 subtests
- `go test ./internal/server/` — running locally; result will be reported before merge

**Negative control, generated from the committed artifact.** Reverting each behaviour independently:

| reverted | tests that went red |
|---|---|
| unconditional `return nil` | total-failure, cancelled, cancelled-precedence |
| summary without `Failed` | always-reports-failures, distinguishes-cancellation |

Critically it **discriminates**: the partial-failure, nothing-to-do and all-skipped cases still *passed* under the revert, because `return nil` satisfies them. A control that turned everything red would only have proved the tests run. Restored afterwards — `git status` clean, full package green.

## Not claimed

The worker-loop `ctx` check is not covered by a test — that would need a `Service` with a mock store, which no test in this package currently constructs. It is a two-line guard verified by build and review only, and I have said so rather than implying the tests cover it.

### DIFF
--- a/changelog.d/20260812_fix_organize_ctx_and_failure_reporting.md
+++ b/changelog.d/20260812_fix_organize_ctx_and_failure_reporting.md
@@ -1,0 +1,31 @@
+### Fixed
+
+#### Organize reported success when every book failed, and could not be cancelled
+
+Three defects on the same path, all of which made an organize run describe itself as
+something other than what happened.
+
+**It always reported success.** `PerformOrganize` ended in an unconditional `return nil`.
+Every book could fail and the caller still saw success, so the operation was recorded as
+succeeded and nothing upstream had any way to learn otherwise. A run now returns an error
+when it was cancelled, or when it failed for every book it attempted. A partial failure
+stays a success on purpose — one failure in three thousand books should not fail the whole
+operation — with the count carried by the summary instead.
+
+**The summary hid failures.** The logged summary listed organized, re-organized,
+already-correct and skipped, but not failed. The failure count existed only in the
+`organize_summary` operation-change row, which nobody reads. A run in which every book
+failed therefore printed `Organize complete: 0 organized, 0 re-organized, 0 already
+correct (stamped), 0 skipped` — indistinguishable from a run that had nothing to do. The
+summary now always states the failure count and the total attempted, and a cancelled run
+says `CANCELED` rather than `complete`.
+
+**Cancellation only half worked.** `organizeBooks` accepted a `context.Context` and never
+read it; cancellation was checked solely in the job feeder. Stopping a run therefore
+stopped new work being queued but let the eight workers drain everything already buffered,
+and context cancellation — an HTTP client disconnecting, or server shutdown — did nothing
+at all. Both the feeder and the worker loop now check `ctx` as well as the operation's own
+cancel flag.
+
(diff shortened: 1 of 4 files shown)


jdfalk:
## Local verification complete

`go test ./internal/server/` — **exit 0, 577.7s**, read from the tool rather than through a pipe.

That was the one outstanding item in the PR body. Full local picture:

| check | result |
|---|---|
| `go build ./internal/organizer/ ./internal/server/` | exit 0 |
| `go vet ./internal/organizer/ ./internal/server/` | exit 0 |
| `go test ./internal/organizer/` | exit 0 — 5 tests, 11 subtests |
| `go test ./internal/server/` | **exit 0, 577.7s** |

The `internal/server` run matters here specifically because this PR changes `library_core_ops.go` — the only caller of `PerformOrganize`, and the place where the new `ErrOrganizeCanceled` branch was added.

PR #2340, opened 2026-08-12
test(abs): seed the conformance fixtures from the oracle, and bound the allowances

jdfalk:
## What

#2337 turned value comparison on and found twelve endpoints that could not meet the oracle.
This closes them: the fake library's two books are now seeded **from the oracle fixture**, and
what genuinely cannot be aligned is named in a **bounded** allowance instead of being left
unchecked.

## The seed was inventing facts

The fake library was built from constants written in the test file. The multi-file book had
six identical 1662-second tracks of 2049–2054 bytes; the real capture has six distinct
durations and files of 11–21 MB. The single-file book was 4096 bytes against a real 120 MB
m4b, and was called `odyssey.m4b` where the oracle says `odyssey_complete.m4b` — a name that
reaches clients twice, through `metadata.filename` and through `trackTitle`'s basename
fallback.

Durations, sizes, filenames, per-track tags and chapter boundaries now come out of the
fixture. Drift is no longer a thing that can happen; it is the same single-source-of-truth
move as deriving the route check from `router.Routes()` in #2335.

Sizes conform via **sparse files**. `mapper.go:225` stats every track and lets the on-disk
size override the recorded one, so `metadata.size` can only match if the file really is that
long — ~200 MB per run, written as holes at no disk cost. Verified safe for the Range
assertions first: `play_test.go:263` compares a suffix range against `full.Body.Bytes()`'s own
tail, not against a byte pattern, so zeroes satisfy it identically.

## Bounded allowances, and the guard that made them necessary

A blanket allowance at `media.duration` accepts **any** duration there forever. A suite full
of those reports green while saying nothing. So an allowance now states the widest gap its
stated cause can produce — 0.5 s for one rounded track, 3.0 s for a six-track aggregate — and
anything wider fails as a different defect:

```
value_mismatch at book[0].libraryItem.media.tracks[0].chapters[0].end:
  want 1386.057, got 1662.5 — gap 276.443 EXCEEDS the 3 allowed by "*end"
```

That is not a hypothetical. `sixChapters()` was returning flat 1662.5-second spans where the
oracle's first chapter ends at 1386.057. A blanket "chapter bounds may differ" would have
absorbed a **276-second** error as the known sub-second truncation. Those chapters are now
seeded from the oracle and match exactly — `database.Chapter` holds float64, so no allowance
is needed at all.

The second guard: **an allowance that never fires fails the test.** The old doc comment asked
readers to notice and delete dead weight themselves; nobody was going to. It turned out to be
a design tool rather than a safety net — every "never fired" failure was the suite telling me
a divergence I had predicted did not exist, which is how `search` ended up with its own exact
list instead of inheriting the multi-file book's. I could not have written these lists
correctly by inspection.

## One correction carried over from #2337

That PR's description claimed **two** mapper defects. Only one was real, and it is corrected
here in the changelog fragment and at the call site.

`metaTags.tagTrack` is **not** a defect. I read one diff row (`want 4, got 4/24`) and inferred
a normalization rule. All six oracle tracks read `1/24, 2/24, 3/24, 4, 5/24, absent` — ABS
passes the raw ID3 tag through, track 4's real tag was a bare `4`, track 6 had none, and
`mapper.go:758` does exactly the same. The seed's hardcoded `"%d/24"` manufactured the
difference, and seeding from the oracle removed it. **One row of a diff is one sample.**

## Five production findings, all filed rather than fixed here

Comparing values surfaced these. None is fixed in this PR — per the standing rule that
production changes do not ride along with test-fixture work — and all are in
`todo.d/20260812-bookfile-duration-integer-seconds.md`:

| Finding | Measured |
|---|---|
| **`BookFile.Duration` is an `int`** | 2.431 s lost across six tracks; `startOffset` drifts 2.200 s by track 6, ~0.4 s per boundary. A client seeking by `startOffset` lands further off the deeper into the book it is |
| **`/api/me/sessions` `itemsPerPage`** | `me.go:132` reports `len(out)`, the item count, not the page size. `stats.go:108,133` get it right |
| **`deviceInfo.deviceType`** | never derived from the User-Agent; always `unknown` |
| **`BookFile.BitrateKbps` is int-kbps** | ffprobe's 96208 bps round-trips as 96000 |
| **`publishedYear`** | rendered from `Book.PrintYear` (int), so a raw `"800BC"` tag returns `"800"` |

Owner decisions taken on the first, third and fifth: file the duration schema change rather
than bundle it; keep our track title deliberately ("the file name is stupid"); allow the
hardcoded `timeBase` rather than add an ingest field for a value nothing is known to read.

## Also

`mediaItemId`, `bookId`, `date` and `dayOfWeek` are now volatile in conformance comparisons —
the first two are the same minted sync ids as `libraryItemId` under a different name, the last
two are when a captured session happened. Four permanent "mismatches" no run could satisfy.
Note this is normalizing things that carry **no signal**; nothing that carries signal was
normalized to chase green, which is what §N-2 of the audit warns against.

`assertConformantPending` is deleted. The debt list it existed for is empty, and a helper kept
alive for a debt nobody owes is a softer gate waiting to be reached for.

## Verification

`go vet` exit 0 · `gofmt` clean · full `internal/server/handlers/abs` and
`internal/syncapi/conformance` packages, **no `-run` filter**.

Every allowance in the suite is proven live by the never-fired guard: if any one of them
stopped describing a real divergence, this PR would not be green.

### DIFF
--- a/changelog.d/20260812-abs-conformance-reseed.md
+++ b/changelog.d/20260812-abs-conformance-reseed.md
@@ -1,0 +1,47 @@
+### Changed
+
+- **The ABS conformance fixtures are now seeded from the oracle, so they cannot drift
+  from it.** The fake library's two books were built from constants invented in the test
+  file: six identical 1662-second tracks of 2049–2054 bytes against a real *Odyssey*
+  capture with six distinct durations and 11–21 MB files, and a single-file book of 4096
+  bytes against a real 120 MB m4b. That drift is why twelve assertions could not compare
+  values at all. Per-track durations, sizes, filenames, tags and chapter boundaries now
+  come out of the fixture itself.
+
+  File sizes conform via sparse files: `mapper.go:225` stats every track and lets the
+  on-disk size override the recorded one, so `metadata.size` can only match if the file
+  really is that long — about 200 MB per run written as holes, costing no disk. Safe for
+  the Range assertions because they compare a suffix range against the response's own
+  tail rather than against a byte pattern.
+
+- **Allowances are now bounded, and an allowance that never fires fails the test.** A
+  blanket exemption at `media.duration` accepts any duration there forever; the day we
+  report half a book's length, a suite full of such exemptions says nothing. A bounded
+  allowance states the widest gap its stated cause can produce — 0.5 s for a rounded
+  track, 3.0 s for a six-track aggregate — and anything wider still fails as a different
+  defect.
+
+  This is not theoretical. The bound immediately caught `sixChapters()` returning flat
+  1662.5-second spans where the oracle's first chapter ends at 1386.057: a **276-second**
+  gap that a blanket "chapter bounds may differ" would have absorbed as the known
+  sub-second truncation. Those chapters are now seeded from the oracle and match exactly,
+  so they need no allowance at all.
+
+### Fixed
+
+- **`mediaItemId`, `bookId`, `date` and `dayOfWeek` are treated as volatile in
+  conformance comparisons.** The first two are the same minted sync ids as
+  `libraryItemId`, reached under a different name from the progress and playback bodies;
+  the last two are when a captured listening session happened. Four permanent
+  "mismatches" that no run could ever satisfy.
+
+### Added
+
+- **Five production findings surfaced by comparing values**, filed rather than fixed
+  alongside test work: `BookFile.Duration` is an `int` (2.431 s lost across six tracks,
+  `startOffset` drifting 2.200 s by track 6, ~0.4 s per track boundary);
+  `/api/me/sessions` reports `itemsPerPage` as the item count rather than the page size
+  where its siblings in `stats.go` get it right; `deviceInfo.deviceType` is never derived
+  from the User-Agent; `BookFile.BitrateKbps` is int-kbps so 96208 bps round-trips as
+  96000; and `publishedYear` renders `Book.PrintYear` as an int, so a raw `"800BC"` tag
+  comes back `"800"`. See `todo.d/20260812-bookfile-duration-integer-seconds.md`.
--- a/changelog.d/20260812-abs-conformance-values.md
+++ b/changelog.d/20260812-abs-conformance-values.md
@@ -18,14 +18,17 @@
   are enumerated in the source rather than in a tracking doc, and the count cannot drift
   upward quietly.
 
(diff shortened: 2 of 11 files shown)


jdfalk:
## Rebased onto current main, re-verified, and one observation recorded

Branch force-pushed after rebasing onto `main` (which now carries the search-pagination fix, the `EmbeddingStore.Close` deadlock fix, and the organize failure-reporting fix).

| check | result |
|---|---|
| `go build ./...` | exit 0 |
| `go vet` (abs + conformance) | exit 0 |
| `go test` (abs + conformance) | **exit 0** — 23.8s / 0.2s |

### A failure I saw once and could not reproduce

The **first** execution of this suite failed with six `value_mismatch` findings:

```
book[1].libraryItem.media.duration: want 9975.431111000002, got 9976
    — gap 0.5688889999983076 EXCEEDS the 0.5 allowed by "book[].libraryItem.media.duration"
book[1].libraryItem.media.tracks[1..5].startOffset: want 1386.057143, got 1386  (etc.)
```

**It has not reproduced in 27 runs since**, across three invocation shapes:

| invocation | runs | failures |
|---|---|---|
| `-run TestSearch_ConformsToOracle` alone | 20 | 0 |
| full `./internal/server/handlers/abs/` | 2 | 0 |
| the exact original two-package command | 5 | 0 |

**The part that does not add up:** the message reports a **0.5** bound applied to `book[].libraryItem.media.duration`, but `browse_test.go` declares that key `Within: 3.0`, and `tracks[].startOffset` `Within: 3.0` too. On the code as written, none of those six gaps (max 1.02) should have been findings at all.

Added `todo.d/20260812-abs-conformance-onefailure-unreproduced.md` with the full output and the lead — whether allowance selection ranges over the `map[string]allowance` and takes the first match, which would make the chosen bound depend on Go's randomised map iteration order, i.e. flaky by construction. **Recorded as a lead, not a finding**: 20 isolated runs did not produce it, so I have not found the mechanism and am not claiming to have.

It is worth keeping visible even though the suite is green: the bounded-allowance design's whole guarantee is that a bound states the widest gap its stated cause can produce. If which bound applies is nondeterministic, a too-wide bound could silently accept a real divergence on some runs — a worse failure than the one observed, because it is invisible.

PR #2348, opened 2026-08-12
PROBE (do not merge): verify the golang.org/x/* license policy actually works

jdfalk:
**Do not merge. This closes as soon as Dependency Review reports.**

#2344 added `allow-dependencies-licenses` purls so Dependency Review stops rejecting `golang.org/x/*` on the compound expression `BSD-3-Clause AND LicenseRef-scancode-google-patent-license-golang`. It shipped **unverified**: #2305, the PR that would have exercised it, was merged at 22:52 — 19 minutes before the fix landed at 23:11 — so `@dependabot rebase` came back "already been merged".

Dependency Review only inspects **changed** dependencies, so any PR that touches no deps passes trivially and proves nothing. This bumps `golang.org/x/arch v0.29.0 → v0.30.0` — the exact module that triggered the original block — which is the only way to learn whether `pkg:golang/golang.org/x/arch` parses the way the action expects. My pre-merge check was a regex for well-formedness against my own assumption, not against the action's parser.

Expected: **Dependency Review passes.** If it fails, the purl form is wrong and #2344 needs a follow-up.


jdfalk:
Verified: **Dependency Review passed** on `golang.org/x/arch v0.29.0 → v0.30.0` — the exact module and change shape that failed before #2344. The `pkg:golang/golang.org/x/arch` purl form parses as intended, so the license policy works. Closing; the dep bump itself was never the point.

PR #2355, opened 2026-08-13
fix(scanner): a failed metadata pass no longer files books under placeholder names

jdfalk:
## What

`scanFolder` logged `"Failed to process books"` and then **fell straight through into auto-organize**.

Those books never got their metadata extracted, so title/author were whatever the scan started with — frequently empty. Organizing them expands the naming pattern over blank fields, sending every one to the same degenerate destination, where all but the first fail as "target already occupied".

## Production evidence — 4h15m scan of 2026-08-11

| signal | count |
|---|---|
| this branch firing with `context canceled` | 2,406 |
| `safeRename refusing to overwrite existing destination` | 7,561 |
| organize-collision candidates | 3,481 |
| books collided on **one** path | **848** |

That path being `Unknown Author/Unknown Title/Unknown Title - Unknown Author.mp3`.

## Also fixed: the folder was counted as a success

The old code logged and then `return nil`. A folder whose metadata pass failed **outright** was indistinguishable from one that worked — the same silent-failure shape as the rest of the sweep.

## Blast radius

One folder, not the scan. The caller handles it:

```go
err := ss.scanFolder(...)   // service.go:195
if err != nil {
    log.Error("Error scanning folder %s: %v", folderPath, err)
    continue
}
```

That is the behaviour we want: a cancelled scan should skip its remaining folders rather than file them under placeholder names.

## Verification

`go build ./...`, `go vet ./internal/scanner/` and `go test ./internal/scanner/` all pass.

This was uncommitted work found in a worktree during a sweep for unlanded changes.

### DIFF
--- a/changelog.d/20260812-scan-metadata-failure-no-organize.md
+++ b/changelog.d/20260812-scan-metadata-failure-no-organize.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- A scan whose metadata pass failed logged the error and then **carried on into
+  auto-organize anyway**. Those books still held whatever title/author the scan
+  started with — frequently empty — so organizing them expanded the naming pattern
+  over blank fields and sent them all to the same degenerate path. In the 4h15m
+  production scan of 2026-08-11, 848 books collided on
+  `Unknown Author/Unknown Title/Unknown Title - Unknown Author.mp3`. The failure now
+  skips auto-organize for that folder and returns, so the folder is also no longer
+  counted as a successful scan.
--- a/changelog.d/20260812_fix_scan_ctx_and_unprocessed_organize.md
+++ b/changelog.d/20260812_fix_scan_ctx_and_unprocessed_organize.md
@@ -1,0 +1,21 @@
+### Fixed
+
+#### A cancelled scan kept going, and filed books whose details were never read
+
+Two faults on the scan path, found while measuring why a full scan takes over four hours.
+
+**Cancelling a scan did not stop it.** The scan checked its own stop button but never
+the request's cancellation signal. When that signal fired, the scan carried on into
+every remaining folder — 2,406 of them in a single production run — failing to read
+details for each one and then reporting overall success.
+
+**Books were filed even when their details could not be read.** When reading a folder's
+audiobook details failed, the failure was written to the log and then the books were
+organized anyway. A book whose title and author were never read has no title and no
+author, so the naming pattern produced the same placeholder name for all of them and
+they were all filed to the same destination — where every one after the first failed as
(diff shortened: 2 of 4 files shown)


jdfalk:
## Correction to this PR's description — the 126,928 figure was misattributed

The description ends with:

> The biggest single contributor by log volume is `RecomputeBookAggregates updated` at **126,928** calls in one scan — roughly two per file.

**"in one scan" is not established, and the "two per file" reading is wrong.** Both were inferred from co-occurrence in the same time window. Correcting on the record:

### What IS established

- `RecomputeBookAggregates updated` fired **126,928 times** between 06:30 and 10:46 on 2026-08-11, across **5,595 distinct books**.
- The distribution is not "two per file" — it is heavily skewed and quadratic in shape. One book (`01KZR90RRM…`) logged **1,189** recomputes; the next three logged 616, 416 and 400.
- `RecomputeBookAggregates` re-reads **all** of a book's files on each call and rewrites the book row. So N sequential inserts for one book cost 1+2+…+N file-record reads. Summing over the observed distribution: **5,430,858 file-record reads where 126,928 would do — 42.8× amplification** — plus 126,928 book-row writes where 5,595 would do.
- The trigger is real and in the code: `CreateBookFile`, `UpdateBookFile`, `DeleteBookFile`, `DeleteBookFilesForBook` and `DeleteBookFilesByIDs` each call `notifyBookFileChange` → a full `RecomputeBookAggregates`.

### What is NOT established

**Which operation issued those calls.** The scanner's file-write path is `BatchUpsertBookFiles` (`internal/scanner/scanner.go:1544`), and `BatchUpsertBookFiles` **never calls `notifyBookFileChange` at all**. So the recomputes did not come from the scan writing book files, and attributing them to the scan was wrong. `RecomputeBookAggregates updated` lines carry **no `opID`** — 0 of 126,928 — so the log cannot answer who caused them. Candidates co-occurring in the window include the organize path (10,369 re-organizes) and the `CreateBookFile` mint path (27,018 `book_file PID uniqueness: transferred to new row`), but I have not traced it.

### The genuinely useful finding underneath

The two write paths are **inconsistent, and neither is right**:

| path | recomputes per book |
|---|---|
| `CreateBookFile` / `UpsertBookFile` (per file) | **N** — O(N²) reads |
| `BatchUpsertBookFiles` | **0** — aggregates left stale |

The correct behaviour is **once per affected book** in both. A coalescing scope (collect touched book IDs, flush one recompute each at the end) fixes both at once. Scope it per book rather than per scan, so aggregates do not sit stale for the length of a 4h15m run.

Not implemented here — it is a change to the core DB write path and deserves its own PR with the attribution question answered first, so the fix can be measured against a known workload rather than a guessed one.

PR #2363, opened 2026-08-13
fix(filters): reject empty filter values instead of matching every book

jdfalk:
## The bug

`fieldMatchesValue` ends in:

```go
return strings.Contains(strings.ToLower(bookValue), strings.ToLower(value))
```

**`strings.Contains(anything, "")` is always true in Go**, so a filter whose value is empty
constrains nothing — it matches every book.

Measured live on production 2026-08-13 against the current deployed build. The filter is
otherwise healthy; this is specific to the empty value:

| `filters=[{"field":"title","value":X}]` | total returned |
|---|---|
| `X = ""` | **63,870 — the entire library** |
| `X = "zzqqxx"` | 0 |
| `X = "Skills"` | 25 |

## Why this is more than a confusing list

`FieldFilters` also flow into `Server.resolveFilterToBookIDs`
(`internal/server/metadata_ops.go:458`), which resolves a `FilterSpec` into **concrete book
IDs for background operations** at `limit=100000` — used by
`metadata_batch_candidates.go:59` and the bulk metadata fetch op.

An empty value there silently retargets a scoped job at the whole library. That is the
**base64 op-params defect (#2309) one level down**: same shape, same whole-library default,
different entry point.

## Fixed in all three layers that were silent

| Layer | File | Behaviour |
|---|---|---|
| HTTP boundary | `handlers/audiobooks/handler.go` | 400 naming the offending field |
| Background op | `metadata_ops.go` | `resolveFilterToBookIDs` returns an error — params arrive already deserialized from the queue, so there is no HTTP boundary here |
| Matcher | `service_filtering.go` | `matchesFieldFilters` fails **closed** |

Layer 3 fails closed rather than open because **matching nothing is visibly wrong and
harmless; matching everything is invisibly wrong and, on the op path, destructive.**
`Negated` is deliberately not consulted — neither `f == ""` nor `f != ""` is a constraint a
caller can have meant, and letting them disagree would be its own trap.

No in-repo code constructs an empty-value filter (checked: the list warmer's ~20
constructions all carry real values), so layer 3 only ever fires on input the boundary
should already have rejected.

## Verification

- `go build ./...` — exit 0
- `go vet ./internal/audiobooks/ ./internal/server/handlers/audiobooks/` — exit 0
- `go test ./internal/audiobooks/ -count=1` — **whole package**, exit 0 (39s)

**Negative control** — removing only the layer-3 guard from the committed file:

| Test | Under negative control |
|---|---|
| `TestEmptyFilterValueDoesNotMatchEveryBook` | FAIL |
| `TestEmptyValueRejectsTheWholeFilterSet` | FAIL |
| `TestEmptyFilterValueFailsClosedEvenWhenNegated` | PASS (negation already excluded it by a different route) |
| `TestNonEmptyFiltersStillWork` (6 cases) | PASS |
| `TestFirstEmptyFilterValue` (3 cases) | PASS |

Discriminating, not blanket-red. `TestNonEmptyFiltersStillWork` is the control that matters:
without it, the two failing tests would pass equally well against a matcher that rejected
*everything*, which would be a worse bug than the one being fixed.

## Not addressed

- **The frontend was not traced.** Whatever sends an empty value now gets a 400 instead of
  the whole library — the safe direction, but the UI path producing it is unidentified, so
  a user may see a raw error.
- **Stored filters were not checked.** A saved smart playlist containing an empty-value
  filter would now return nothing instead of everything.

### DIFF
--- a/changelog.d/20260813_fixed_empty_filter_matches_whole_library.md
+++ b/changelog.d/20260813_fixed_empty_filter_matches_whole_library.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- A filter with an empty value silently matched every book instead of narrowing anything.
+  Filtering the library by, say, title with the box left empty returned the entire library
+  rather than an error or an empty result — and the same filters are used to pick which
+  books a background metadata operation runs against, so an empty value could quietly point
+  a targeted job at the whole collection. Empty filter values are now rejected with a clear
+  message explaining that omitting the filter is how you ask for everything.
--- a/internal/audiobooks/empty_field_filter_test.go
+++ b/internal/audiobooks/empty_field_filter_test.go
@@ -1,0 +1,128 @@
+// file: internal/audiobooks/empty_field_filter_test.go
+// version: 1.0.0
+// guid: 4e8b1a37-9d52-4c06-b3f8-27a1d0e5c964
+// last-edited: 2026-08-13
+
+package audiobooks
+
+import (
+	"testing"
+
+	"github.com/falkcorp/audiobook-organizer/internal/database"
+)
+
+// emptyFilterTestBook builds a book with a known title/narrator so the tests
+// below can distinguish "matched because the filter is satisfied" from "matched
+// because the filter constrained nothing".
+func emptyFilterTestBook(title string) database.Book {
+	n := "Some Narrator"
+	return database.Book{
+		ID:       "b1",
+		Title:    title,
+		Narrator: &n,
+	}
+}
+
+// TestEmptyFilterValueDoesNotMatchEveryBook is the regression for the defect.
+//
+// strings.Contains(anything, "") is true in Go, so before the fix an empty
+// filter value silently matched every book. Measured on production 2026-08-13:
+// title="" returned all 63,870 books while title="zzqqxx" returned 0 — the
+// filter worked, and only the empty value widened it to everything.
+func TestEmptyFilterValueDoesNotMatchEveryBook(t *testing.T) {
+	book := emptyFilterTestBook("The Hobbit")
+
+	if matchesFieldFilters(book, []FieldFilter{{Field: "title", Value: ""}}) {
+		t.Error("an empty filter value matched a book; it must fail closed, because " +
+			"strings.Contains(x, \"\") is always true and would match the whole library")
+	}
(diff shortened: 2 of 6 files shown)


jdfalk:
Follow-up on the "frontend was not traced" caveat — I traced it, and the answer is **no UI path can produce an empty filter value**, so the new 400 will not break the app:

| Construction site | Guard |
|---|---|
| `web/src/utils/searchParser.ts:130` | `const trimmedValue = fieldMatch.value.trim(); if (trimmedValue) { ... }` — empty and whitespace-only values are dropped before the push |
| `web/src/pages/Library.tsx:802-805` | `if (filters.author)` / `.series` / `.genre` / `.language` — truthiness, so `""` is skipped |
| `web/src/pages/Library.tsx:808` | passes through values that already came from the parser above |

Those are all the `fieldFilters.push` sites in `web/src` (the other `{ field: ... }` grep hits are `MaintenanceSettingsSection.tsx` task toggles, an unrelated shape).

So the production exposure measured in the PR body — `title=""` returning all 63,870 books — was reachable through **direct API calls and the background-operation params path**, not through the library UI. That does not reduce the severity of the `resolveFilterToBookIDs` case, which is the one that could retarget a scoped metadata job at the whole library; it does mean the boundary rejection is a pure safety net rather than a behaviour change for the app.

Still not checked: whether any **stored** smart playlist or saved filter contains an empty value. Such a filter would now return nothing instead of everything.

jdfalk:
Second caveat closed — **stored smart playlists are not affected.**

`matchesFieldFilters` has exactly one non-test caller outside its own file: `internal/audiobooks/service_query.go:382` (via `matchesFieldFiltersWithStrippedFallback`), which is the library list query path. Smart playlists evaluate through a separate matcher, `internal/search/per_user_match.go`, which never calls it.

So the layer-3 fail-closed change has a blast radius of exactly one path — the library list — and a saved playlist containing an empty-value filter would keep whatever behaviour it has today rather than flipping from "everything" to "nothing".

Both "not addressed" items from the PR body are now resolved:
1. Frontend — no UI path can emit an empty value (comment above).
2. Stored filters — playlists use a different evaluator, unaffected (this comment).

PR #2407, opened 2026-08-14
fix(database): exclude soft-deleted books from series counts on the Pebble path

jdfalk:
## What

`GetAllSeriesBookCounts` and `GetAllSeriesFileCounts` have two implementations and they disagreed. `MemStore` has always skipped soft-deleted books; the Pebble key-scans counted them — so a series reported its trashed books in its book count, and their audio files in its file count.

## Blast radius

memdb is the production default, so the visible effect was confined to:
- **cold start** — the window after process start but before warmup publishes the in-memory index, and
- any deployment running with `UseMemDB=false`.

In that window a series' counts read high and then silently corrected themselves once warmup completed, with nothing logged to explain the change. **This is a display-aggregate defect only** — both call sites (`filesystem.go` folder `BookCount`, `scanner/service.go` import-path count) assign to a count field. Nothing gates a delete or cleanup on these numbers.

Because prod runs memdb, **prod behaviour is unchanged by this PR** and I have not validated it against prod.

## The phase-1 detail

In `GetAllSeriesFileCounts` the skip goes in **phase 1**, where the `bookID → seriesID` map is built, rather than in the phase-2 `book_file:` loop. A book kept out of that map never has its `BookFile` rows attributed to the series, so one check covers both counts and avoids a second book lookup per file.

## Test

New `TestSeriesCounts_ExcludeTrashOnBothPaths` runs one fixture through both implementations with `UseMemDB` flipped and asserts they agree. Fixture is built so the filter is falsifiable: series A has 2 live + 2 trashed books (forgetting the skip reads 4), series B has no trash at all (a mutation skipping *every* book also fails).

It guards non-vacuity with `require.True(t, p.IsMemReady())` — without that, the `UseMemDB=true` arm can silently run the Pebble path and the test asserts memdb == memdb. Both these methods already gated correctly on `p.UseMemDB && p.mem() != nil`, so unlike `ListBooksByITunesPID` in #2399 the flip genuinely selected two paths here.

**Mutation-verified.** Reverting either half turns the test red on the specific count it governs:

| mutation | result |
|---|---|
| drop trash-skip from `GetAllSeriesBookCounts_Pebble` | FAIL — "series A book count must exclude its 2 trashed books" |
| drop trash-skip from `GetAllSeriesFileCounts` | FAIL — "series A file count must exclude files of trashed books" |
| both restored | PASS |

## Provenance

Found by auditing every dual-implementation store method for soft-delete drift, following #2392/#2399. Fourteen candidates were machine-flagged; ten were false positives and were hand-eliminated by reading the code. Three other methods genuinely drift (`ListBookIDs`, `CountBooksByPathPrefix`, plus two ungated memdb dispatches) — those are a **separate PR** because they needed a caller survey to confirm no consumer treats absence-from-the-list as licence to delete.

### DIFF
--- a/changelog.d/20260814_120000_fix_series_counts_exclude_trash.md
+++ b/changelog.d/20260814_120000_fix_series_counts_exclude_trash.md
@@ -1,0 +1,24 @@
+### Fixed
+
+#### Series book and file counts no longer include trashed books on the Pebble path
+
+`GetAllSeriesBookCounts` and `GetAllSeriesFileCounts` have two implementations,
+and they disagreed. The memdb implementation has always skipped soft-deleted
+books; the Pebble key-scan counted them. A series whose trash held four books
+reported those four in its book count, and their audio files in its file count.
+
+memdb is the production default, so the visible effect was confined to cold
+start — the window after process start but before warmup publishes the in-memory
+index — and to any deployment running with `UseMemDB=false`. In that window a
+series' counts read high and then silently dropped to the correct value once
+warmup completed, with nothing logged to explain the change.
+
+The fix adds the same `bookIsSoftDeleted` skip both scans were missing. In
+`GetAllSeriesFileCounts` the skip goes in phase 1, where the book-to-series map
+is built, rather than in the phase-2 file loop: a book excluded from that map
+never has its `BookFile` rows attributed to the series, so one check covers both.
+
+A new cross-backend conformance test (`TestSeriesCounts_ExcludeTrashOnBothPaths`)
+runs the same fixture through both implementations with `UseMemDB` flipped and
+asserts they agree. Both halves of the fix were mutation-verified — reverting
+either one turns the test red on the specific count it governs.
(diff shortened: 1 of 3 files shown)


jdfalk:
Superseded by #2411, which landed the same two fixes first (`bookIsSoftDeleted` in `GetAllSeriesBookCounts_Pebble` and `GetAllSeriesFileCounts`).

I checked whether my conformance test covered anything `aggregate_count_conformance_test.go` misses. It doesn't. My fixture's distinguishing feature was a second series with no trash, as a control against a mutation that skips *every* book — but #2411's asymmetric 5-live/2-trashed split already fails that mutation (expected 5, would get 0). Landing a second conformance test over the same two functions would be duplicate coverage with no added signal, so I'm closing rather than rebasing.

The `ListBookIDs` and `computeLibraryStats` work from the same audit is unique and continues in #2408, rebased on top of #2411.

Both PRs came out of a sweep of every `MemStore`/`PebbleStore` twin for soft-delete drift — 14 candidates machine-flagged, 4 hand-verified real. #2411 independently found and fixed 2 of the 4.

PR #2492, opened 2026-08-16
fix(itunes): stop stub ops from shadowing their real implementations

jdfalk:
## What

Three iTunes operations have been reporting `completed` in production while doing nothing since 2026-07-17.

`Container.PostInit` runs before `NewServer`'s `opRegistrars` loop (`internal/server/server.go:567` vs `:625`), so an `OperationDef` registered by a plugin always wins an ID collision and the **server's** registration is the one rejected.

`internal/plugins/itunes` registered four stubs. Three collided with working implementations:

| Op | Stub (winner) | Real implementation (rejected) |
|---|---|---|
| `itunes.sync` | `sync.go` → `return nil` | `server.RegisterITunesSyncOp` → `Importer.Sync` |
| `itunes.path-reconcile` | `path_reconcile.go` → `return nil` | `server.RegisterITunesPathReconcileOp` → `Paths.Reconcile` |
| `itunes.path-repair` | `path_repair.go` → `return nil` | `server.RegisterITunesPathRepairOp` → `Repair.Repair` |
| `itunes.position-sync` | `position_sync.go` → `return nil` | *(none — sole registrant)* |

The only evidence was one `WARN msg="op registrar" err="OperationDef already registered"` per boot — **378 lines in 30 days** of production journal — and the registrar loop swallowed it.

`importDef` was already excluded, with a comment describing this exact hazard. It was never applied to its four siblings.

## Changes

- `registeredDefs()` is now an explicit whitelist, split out of `Register` so it can be asserted without an enabled `Service` (`Enabled()` reads an unexported `deps.Config`).
- Every stub `Run` returns an error instead of `nil`. An unimplemented op that returns `nil` is recorded `completed` — a false report, not a placeholder.
- `itunes.position-sync` stays registered: it has no counterpart, and dropping it would make the op vanish rather than fail.

## Tests

`TestStubRuns` previously asserted `if err != nil { t.Fatalf }` — it **locked in the false success it appeared to guard**. Inverted.

Both new guards mutation-tested:
- re-adding the 3 shadowing defs → `TestRegister_OnlyRegistersDefsWithARealRun` fails naming all 3
- reverting one stub to `return nil` → `TestStubRuns_FailInsteadOfReportingSuccess` fails naming it

4/4 pass; `go vet` clean.

## Note on sequencing

`fix/op-registrar-failure-is-fatal` (makes a failed registration refuse to boot) **must merge after this one is deployed** — with the duplicate present it would prevent production from starting.

## Not in scope

`todo.d/20260816-finish-itunes-plugin-migration.md` — finishing or deleting the migration, and wiring `svc.Positions.Sync` (fully implemented, never called by anything; turning it on means real bidirectional writes to user positions on a 63k-book library).

### DIFF
--- a/changelog.d/20260816-itunes-stubs-shadow-real-ops.md
+++ b/changelog.d/20260816-itunes-stubs-shadow-real-ops.md
@@ -1,0 +1,16 @@
+### Fixed
+
+- Three iTunes operations — **Sync**, **Path Reconcile** and **Path Repair** —
+  have been reporting success without doing anything since 2026-07-17. A
+  half-finished refactor left placeholder versions of them in the iTunes plugin,
+  and because plugins load before the rest of the server, the placeholders took
+  over from the working versions. Running any of the three produced a green
+  "completed" row and no work. All three now run their real implementations
+  again. A fourth, **Position Sync**, had no working version to fall back on and
+  now reports an honest failure instead of a false success.
+
+### Changed
+
+- A placeholder operation that has not been implemented yet now fails when run
+  instead of quietly reporting success. Reporting success for work that did not
+  happen is the more expensive failure: nothing looks wrong, so nobody looks.
--- a/internal/plugins/itunes/import.go
+++ b/internal/plugins/itunes/import.go
@@ -8,7 +8,6 @@ package itunes
 import (
 	"context"
 	"encoding/json"
-	"log/slog"
 	"time"
 
 	"github.com/falkcorp/audiobook-organizer/pkg/plugin/sdk"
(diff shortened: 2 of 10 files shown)


jdfalk:
**Measurement correction.** The PR body says these ops were "reporting success without doing anything since 2026-07-17." That accurately describes the code path when one is invoked, but I should be precise about what I did and did not measure:

- **Confirmed:** the stub wins the ID collision (registration order, `server.go:567` vs `:625`), its `Run` returns `nil`, and the real implementation is rejected. 378 `op registrar` WARN lines in 30 days of journal, 3 on the current build.
- **NOT measured:** how many times anyone actually triggered these ops. I checked the journal for the stub's `"op not implemented"` warning and got 0 hits — but that number is meaningless here, because `dbReporter.Log` (`reporter_db.go:366`) buffers to `op_logs_v2` in the database and never reaches slog/journald. Zero journald hits is what you'd see whether it ran a thousand times or never.

So: the defect is confirmed, its blast radius is not. Getting the real invocation count means querying `op_logs_v2` / `operations_v2` for those three def IDs, which the inert `?def_id=` API filter can't currently do either.

The fix doesn't depend on the count.

PR #2495, opened 2026-08-16
fix(scanner): check in during directory walk and honor the configured LLM backend

jdfalk:
Two defects that together killed the 2026-08-16 library rescan. Both were found by reading the journal for the window in which the scan went silent, not by inspection.

## 1. The scan reported nothing while it was working

`ScanDirectoryParallel` ran both of its phases — the `WalkDir` that discovers directories, and the parallel pass that reads each one — without a single `UpdateProgress` call. The stuck-op watchdog kills an operation after `ProgressTimeout` (5m) of silence, so on a large import folder a busy scan is indistinguishable from a hung one.

Production evidence:

```
13:12:11  AI batch parsing disabled ... remaining 1459 books keep filename-derived metadata
13:12:38  scanner multi-file group detected  (busy)
13:14:05  scanner multi-file group detected  (busy)
13:15:18  scanner multi-file group detected  (busy)
13:16:01  scanner multi-file group detected  (busy)
13:17:34  registry: strike recorded  kind=stuck  "no progress for 5m23s (timeout=5m0s)"
13:17:39  op -> interrupted_dropped
13:28:22  library scan failed  err="scan canceled: context canceled"
```

The process was demonstrably working the entire time it was declared stuck — mid-walk of `/mnt/bigdata/books/newbooks`, which holds 17,469 books. Both phases now checkpoint every 20 directories.

## 2. The scanner ignored the configured LLM backend

`scanner.go` built the **cloud** OpenAI parser whenever `EnableAIParsing` was set, passing `enabled=true` and ignoring `ai_backend.llm_mode` entirely. Production had `llm_mode: "disabled"` and a working local Ollama endpoint already configured at `local_base_url` — and the scan still sent every batch to `api.openai.com`, where it got `insufficient_quota / credit_balance_exhausted`.

Parser construction now routes through `EffectiveLLMMode`, handling local / openai / disabled, so it matches the `llmparser` service in `internal/ai/register.go` instead of being a second, divergent copy.

## Testing

`TestScanDirectoryParallel_ChecksInLongEnoughToSurviveTheWatchdog` asserts the **count** of checkpoints, not merely that one happened — a single call at the start or end would satisfy "reports progress" while leaving an arbitrarily long silent stretch in the middle, which is the actual defect.

Mutation-tested: with both checkpoints neutralized the test fails with `got 0 progress checkpoints for 60 directories, want >= 3`; restored, it passes. `go build ./...` and `go vet ./internal/scanner/` clean. (`go vet` caught a real bug during development — the new log line used slog-style key/value pairs on a printf-style logger, which would have silently dropped them.)

## Not addressed here

The AI-batch fix from the previous PR is confirmed working in production — the phase aborted at batch 1 in 16.3s instead of grinding ~25 minutes, matching the real error. But that only relocated the silence to the next phase; this PR fixes that phase. The scan has still never completed end-to-end, and that remains unverified until a run reaches `outcome=completed`.

### DIFF
--- a/changelog.d/20260816-scanner-walk-progress.md
+++ b/changelog.d/20260816-scanner-walk-progress.md
@@ -1,0 +1,15 @@
+### Fixed
+
+- **Library scans no longer get killed while they are still working.** The
+  directory-discovery and directory-scanning phases of a scan reported no
+  progress at all, so on a large import folder the scan looked identical to a
+  hung process and the stuck-operation watchdog killed it after five minutes.
+  The 2026-08-16 rescan died this way mid-walk of a folder holding 17,469
+  books. Both phases now check in every 20 directories.
+
+- **The scanner honors the configured LLM backend.** It previously built the
+  cloud OpenAI parser whenever AI parsing was enabled, ignoring
+  `ai_backend.llm_mode` entirely — so a deployment configured for a local
+  Ollama endpoint, or for no LLM at all, still sent every scan batch to
+  `api.openai.com`. Local, OpenAI, and disabled modes are now all respected,
+  matching the `llmparser` service registration.
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1,5 +1,5 @@
 // file: internal/config/config.go
-// version: 1.78.0
+// version: 1.79.0
 // guid: 7b8c9d0e-1f2a-3b4c-5d6e-7f8a9b0c1d2e
 // last-edited: 2026-08-15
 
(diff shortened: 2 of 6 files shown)


jdfalk:
## Follow-up: per-directory checkpoints were the wrong unit

Raised by the user, and correct — the first version of this fix counted **directories**, which is only safe if a directory is bounded work. It isn't: it can hold one file or 17,469.

A library kept as **one flat folder** has exactly one directory, so:
- the discovery walk reports nothing (`1 % 20 != 0`)
- the scan phase reports once, via the `|| n == len(dirs)` clause — *after* all the work

The whole scan is a single silent stretch and the watchdog kills it. Same failure, different library layout.

### Fix

`groupFilesIntoBooks` now takes an optional per-file callback, invoked from **both** of its unbounded tag-reading passes:
- the multi-file-group detection loop (`quickReadMultiFileInfo` per file)
- the mixed-directory album-grouping loop (`quickReadAlbum` per file) — the one a flat library actually hits, since a folder of unrelated books fails the group check

`ScanProgressEvery` is exposed in config for operators on slow storage; `0` keeps the default of 20.

### Verification

New test `TestScanDirectoryParallel_ReportsInsideOneHugeDirectory` uses `dirCount == 1`, which no directory-counting implementation can pass at any interval.

Mutation-tested — with the per-file callback removed:

```
got 1 checkpoints scanning 60 files in a SINGLE directory, want >= 3
```

Restored: passes. `go build ./...` and `go vet` clean.

PR #2498, opened 2026-08-16
feat(collections): implement the ABS collections surface, static and dynamic

jdfalk:
Closes the user-reported "I can't create collections" and "collections list is empty".

## What was broken, measured

Both halves were unimplemented, in different ways:

| Route | Before |
|---|---|
| `GET /api/libraries/:libraryId/collections` | `h.EmptyPage` — a stub returning a valid, **permanently empty** `Page<T>` |
| `POST /api/collections` | no route at all |

Production journal, five attempts in two seconds from one client — a user pressing Create repeatedly:

```
15:45:10 WARN http request method=POST path=/api/collections status=404
              message="endpoint not found"
```

The list stub was the more dangerous half, and the same defect class as #2496: an empty 200 is indistinguishable from "you have no collections", so nothing errors and nothing logs.

## What this adds

List, detail, create, update, delete, single-book add/remove; the Pebble store layer; and the `collections.manage` permission atom. **Static and dynamic** collections both supported.

## Design decisions the obvious implementation gets wrong

- **Collections are server-wide.** `playlists.go` 404s a playlist whose `CreatedByUserID` isn't the caller. This file is modelled on that one, so copying the check was the likely mistake — it would hide most collections from most users *while looking like a security fix*. `CreatedByUserID` is attribution only.
- **One permission gate, not admin-OR-permission.** `adminPermissions()` is `All()` and `SeedRoles` recomputes existing roles every boot, so admins acquire `collections.manage` with no backfill. A second branch would be dead code.
- **Sync ids are translated.** Clients hold 36-char sync ids; books are 26-char ULIDs. Storing them verbatim would create successfully, return 200, and render empty forever.
- **Dynamic collections serve materialized members** on read (mirroring `playlistDTO`), so a read path never depends on Bleve being open. Their membership isn't editable here — accepting a book list would store a set the next evaluation discards (409).
- **Books are full LibraryItems; `numBooks` is `len(books)` served.** Both #2496 lessons: one undecodable entry discards the whole response in a typed client, and a row reporting a count it cannot list forces the client to guess.

## Namespace move

`/api/collections` moves out of `absUnimplementedNamespaces` into `absReservedPaths` + `absReservedPathPrefixes`. Routing is unchanged by the move (`absReservedPath()` matched either list), but the old placement would have left a list titled "endpoints we do NOT implement" naming one we do — and `TestUnimplementedABSNamespacesAre404NotRedirect` derives its assertions from it, so it would have demanded a 404 from the route that now answers.

A subtree reservation is safe here where it was a live bug for playlists: there is no `/api/v1/collections` twin for a wide reservation to capture. With the ABS surface disabled these paths 404 exactly as before.

## Verification — every test mutation-tested

A green test proves nothing until it has been watched going red. Each behaviour was reverted and the test confirmed to fail:

| # | Mutation | Test that caught it |
|---|---|---|
| 1 | Delete the write gate | `CreateRequiresPermission` |
| 2 | Add the playlist ownership filter | `AreVisibleToEveryUser` |
| 3 | Store client sync ids verbatim | `SyncIDsAreTranslatedToBookIDs` |
| 4 | `numBooks` counts stored, not served | `NumBooksMatchesBooksServed` |
| 5 | Dynamic reads `BookIDs` | `DynamicServesMaterializedMembers` |
| 6 | Restore the `EmptyPage` stub | `CreateThenListRoundTrip` |
| 7 | Allow editing dynamic membership | `DynamicMembershipIsNotEditable` |

Mutation 4 initially "survived" — the sed pattern missed gofmt's alignment, so the edit was a no-op. Re-run against the real text, it failed as required (`numBooks=2 but 1 books served`). A no-op mutation is indistinguishable from a surviving mutant unless you check the file.

`TestCollections_ReadingIsNotGated` is the negative control: without it, a handler that 403'd *every* collection route would pass the gating test while making the feature unusable.

Full ABS package: `ok 30.1s`. Namespace/redirect tests: `ok`.


jdfalk:
## Second commit: the native API — where dynamic collections actually get created

The first commit made the app work. This one makes the *feature* complete, and closes a gap I'd otherwise have shipped quietly:

**Audiobookshelf has no concept of a query-backed collection.** Its create payload has nowhere to put a query, so everything created through the app is static. Without native routes, "collections support static and dynamic" would have been true of the storage layer and **false of the product** — the store would accept a dynamic collection that no caller could construct.

Adds list / get / create / update / delete / materialize under `/api/v1/collections`.

### Rules

- **Every write is gated** on `collections.manage` — unlike the playlist block directly above it in `wire_library_routes.go`, where only the list carries `s.perm` and POST/PUT/DELETE carry nothing. Not an inconsistency to tidy later: collections are server-wide, so an ungated write lets any authenticated user rewrite what everyone else sees. Both surfaces enforce it; one alone would be decorative, since anyone refused on one could call the other.
- **Each field belongs to exactly one type, and the mismatch is rejected rather than ignored.** `book_ids` on a dynamic collection, or a `query` on a static one, would otherwise be a write that returns 200 and changes nothing.
- **A dynamic collection survives an unavailable search index.** It's evaluated once at creation so it's populated the first time anyone looks, but a failed evaluation does not fail the create — refusing it because the index is closed would discard the user's input for an unrelated reason.
- **The query engine is reused** from `internal/playlist`, not reimplemented. A private copy would drift, and a query that worked in a smart playlist but failed in a dynamic collection would be indistinguishable from a bad query.

### The gate test does not use `setupTestServer`

That fixture leaves `EnableAuth` false, and **`s.perm()` returns a pass-through no-op when auth is disabled** — so a gating test written against the production router there would pass with the gate deleted. `RequirePermission` has a second bypass for the same reason (`CountUsers()==0`). The test mounts `servermiddleware.RequirePermission` directly over a store seeded with a user, and asserts the count is positive first so neither bypass can make it vacuous. Precedent: `newAdminInvalidateRouter`.

`TestCollectionWrites_RequirePermission/with the permission` is the negative control — without it, a router that refused *everything* would pass the refusal block.

### Mutations (4 more, all killed, each verified applied)

| # | Mutation | Caught by |
|---|---|---|
| N1 | Allow a dynamic collection with no query | `CollectionCreateValidation` |
| N2 | Ignore `book_ids` on dynamic instead of rejecting | `DynamicMembershipIsNotEditableNatively` |
| N3 | Scope the list to the calling user | `CollectionsAreNotScopedToTheCallingUser` |
| N4 | Fail the create when evaluation fails | `CreateDynamicCollectionSurvivesAnUnavailableIndex` |

Eleven mutations across both commits, all killed. `internal/server/handlers`: `ok 2.1s`.

PR #2503, opened 2026-08-17
refactor(maintenance): narrow the 19 job helpers beneath MaintenanceJob.Run

jdfalk:
First implementation step from `docs/audits/2026-08-16-store-interface-decomposition.md` (#2501).

## Why this layer

`MaintenanceJob.Run` (`internal/maintenance/job.go:70`) takes the full `database.Store`, and an interface method's parameter type is fixed for every implementer — **31 job types implement it**. That single signature is why `internal/maintenance/jobs` holds the largest concentration of wide-store declarations in the codebase (59 of the 338 AST-counted total).

**`Run` is deliberately untouched here.** Re-typing it is a 31-file atomic edit that cannot be split across PRs. But the interface dictates `Run` and *nothing else* — the free helpers beneath it chose `database.Store` with nothing compelling them, and that is where the volume is.

## What changed

19 free helpers now take narrow slices declaring exactly the methods they use (1–4 each, out of 398). Targets were **derived from actual usage** by a `go/types` analyzer, not guessed.

| helper | was | now | methods |
|---|---|---|---|
| `deleteOldOperations` | `database.Store` | `retentionOperationStore` | 2 |
| `deleteDeadPrefixes` | `database.Store` | `retentionKVStore` | 2 |
| `deleteStaleOperationState` | `database.Store` | `retentionOpStateStore` | 3 |
| `is/setDeadPrefixSweepDone` | `database.Store` | `retentionFlagStore` | 2 |
| `rbnrApplyFix`, `vgUnlinkOutliers` | `database.Store` | `bookMutator` | 2 |
| `ddSoftDeleteBook` | `database.Store` | `bookSoftDeleter` | 3 |
| `csUnlinkAndDeleteSeries` | `database.Store` | `seriesUnlinker` | 3 |
| `csMergeSeriesGroup` | `database.Store` | `seriesMerger` | 4 |
| …and 9 more | | | 1–2 each |

## Explicit method lists, not embedded sub-interfaces

Embedding is *not* narrowing. `database.BookStore` alone is **51 methods**, so a helper needing two would still declare 51. The measured cost of that mistake is already in-tree: `internal/reconcile` composes four sub-interfaces for **115 declared methods and uses 11**.

Each slice carries `var _ Slice = (*database.PebbleStore)(nil)` so signature drift is a build error rather than a call-site surprise.

## Why this is safe

Narrowing a parameter is **monotone**. `database.Store` is composed *purely* of embedded sub-interfaces and declares no methods of its own, so anything satisfying `Store` — `*PebbleStore`, and the hand-written `*database.MockStore` — satisfies every slice too. Changing a parameter type is a *widening* of what the call site accepts.

**Zero call sites and zero tests changed.** The diff is the 19 signatures, `store_slices.go`, and header bumps.

## Verification

- `go build ./...` — clean
- `go vet ./...` — clean, **full tree, never scoped** (the April 18 / PR #394 post-mortem records that scoped vet misses test-file breakage)
- `go test ./internal/maintenance/...` — passing

## Not in scope

The 3 remaining non-`Run` helpers in this package propagate the store to a still-wide callee and must be done in topological order; they are now unblocked by this PR. `internal/plugins/maintenance` has 16 more.

### DIFF
--- a/changelog.d/20260816_220000_narrow_maintenance_job_helpers.md
+++ b/changelog.d/20260816_220000_narrow_maintenance_job_helpers.md
@@ -1,0 +1,18 @@
+### Changed
+
+- `internal/maintenance/jobs` — the 19 free helper functions beneath
+  `MaintenanceJob.Run` now take narrow, explicitly-declared store slices instead of the full
+  `database.Store` (398 methods). Each helper uses 1-4 methods; the new slices in
+  `store_slices.go` declare exactly those, with `var _ Slice = (*database.PebbleStore)(nil)`
+  assertions so signature drift is a build error rather than a call-site surprise.
+
+  `MaintenanceJob.Run` itself is unchanged — an interface method's parameter type is fixed for
+  all 31 implementers. Narrowing the layer beneath it captures the volume without a 31-file
+  atomic edit.
+
+  Narrowing a parameter is monotone (`Store` is composed purely of embedded sub-interfaces,
+  so anything satisfying `Store` satisfies every slice), so **no call site and no test
+  changed**. Verified: `go build ./...`, full-tree `go vet ./...`, and
+  `go test ./internal/maintenance/...` all clean.
+
+  Rationale and the measured baseline: `docs/audits/2026-08-16-store-interface-decomposition.md`.
--- a/changelog.d/20260816_220000_narrow_maintenance_job_helpers.md
+++ b/changelog.d/20260816_220000_narrow_maintenance_job_helpers.md
@@ -1,18 +1,20 @@
 ### Changed
 
-- `internal/maintenance/jobs` — the 19 free helper functions beneath
-  `MaintenanceJob.Run` now take narrow, explicitly-declared store slices instead of the full
-  `database.Store` (398 methods). Each helper uses 1-4 methods; the new slices in
+- `internal/maintenance/jobs` — 19 declarations beneath `MaintenanceJob.Run` (16 free helper
+  functions and 3 methods) now take narrow, explicitly-declared store slices instead of the full
+  `database.Store` (398 methods). Each uses 1-4 methods; the new slices in
   `store_slices.go` declare exactly those, with `var _ Slice = (*database.PebbleStore)(nil)`
   assertions so signature drift is a build error rather than a call-site surprise.
 
   `MaintenanceJob.Run` itself is unchanged — an interface method's parameter type is fixed for
-  all 31 implementers. Narrowing the layer beneath it captures the volume without a 31-file
-  atomic edit.
+  all **37** implementers (measured: 37 `Run` receivers, 37 files, 37 non-test
+  `maintenance.Register` calls, one each). Narrowing the layer beneath it captures the volume
+  without a 37-file atomic edit.
 
   Narrowing a parameter is monotone (`Store` is composed purely of embedded sub-interfaces,
-  so anything satisfying `Store` satisfies every slice), so **no call site and no test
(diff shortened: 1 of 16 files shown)


jdfalk:
## Added: Option C as a worked example (`d6970f28`)

Both patterns now sit side by side in the **same file**, so this PR is a comparison of two implementations rather than of one implementation and a sketch.

- **B (narrow interface)** — `deleteDeadPrefixes`, `deleteStaleOperationState`, `isDeadPrefixSweepDone`, `setDeadPrefixSweepDone`, + 15 others
- **C (split pure decision from I/O)** — `deleteOldOperations`

### The split

```go
operationsOlderThan(ops []database.Operation, cutoff time.Time) []string  // pure: no store, ctx or dryRun
expiredOperationIDs(ctx, store operationLister, cutoff)                   // read: 1 method
deleteOperations(ctx, store operationDeleter, ids)                        // write: 1 method
```

`Run` branches on `dryRun` and either reports `len(expired)` or calls `deleteOperations`.

| | B | C |
|---|---|---|
| Declared store surface | 398 → 2 | 398 → 1 + 1 + **0** |
| `dryRun` below `Run` | still threaded | **gone** |
| Diff | **1 line** + 4-line interface | ~80 lines, 5 files |
| Call sites changed | **0** | 1 + 5 test sites |
| Boundary case assertable | only via a fake | **one line, slice literal** |
| Mechanically sweepable | **yes** | no — judgement per site |

### Two findings worth keeping

**The existing test was fake, and only C could expose it.** `TestRetentionBoundaryLogic` asserted `tc.opTime.Before(cutoffTime) == tc.shouldDel` — comparing `time.Before` against itself, never reaching production code. It *couldn't*, because the decision was welded to a store call. B leaves it exactly as-is: a green test asserting nothing in a file that looks well covered. Replaced by a 5-case table over slice literals.

**C produces D rather than competing with it.** Once the decision is out, both I/O halves need exactly one method — the 2-method `retentionOperationStore` became `operationLister` and `operationDeleter`.

### The cost, stated plainly

**C moved an invariant into `Run`** — the one function here that can't be narrowed and is hardest to test. One function used to return both the dry-run count and the real count, so "never report more deletions than you made" held structurally. Now nothing forces them to agree.

Re-pinned deliberately in the same commit: `TestDeleteOperations_CountMatchesInput`, `TestDeleteOperations_PartialCountOnError`, and `TestJobRun_DryRunDeletesNoOperations` (Run-level, real Pebble store). **Budget for this whenever C is applied** — a split that quietly drops an invariant into `Run` looks like a clean refactor in review.

Also: the listing call deliberately stayed *inside* `expiredOperationIDs`. Hoisting it into `Run` would have left the single-pass guarantee testable only through `Run`, which takes the full `database.Store`.

### Verification

`go build ./...` + full-tree `go vet ./...` clean · `go test ./internal/maintenance/...` ok · **mutation-tested 4/4 killed**:

| Mutant | Killed by |
|---|---|
| `Before(cutoff)` → `!After(cutoff)` | `TestOperationsOlderThan` |
| `deleteOperations` returns `len(ids)` on error | `TestDeleteOperations_PartialCountOnError` |
| `Run`'s `if dryRun` → `if false` | `TestJobRun_DryRunDeletesNoOperations` |
| `ListOperations(0,0)` → `(500,0)` | `TestExpiredOperationIDs_ScansOnce` |

### Correction to this PR's own description

Measured A/B with the `go/types` analyzer (details in #2505): this PR narrows **19 declarations = 16 free helpers + 3 methods**, not "19 of the 35 free helpers". **24 remain**, not 19.

PR #2521, opened 2026-08-17
refactor(plugins/maintenance): narrow 19 helpers off the 398-method store

jdfalk:
## What

19 helpers in `internal/plugins/maintenance` now take narrow, explicitly-declared store slices instead of the full `database.Store` (398 methods). New `store_slices.go` declares 15 interfaces of 1–11 methods each.

This is the `OK`-tier (no-propagation) half of the remaining narrowing work, and the sibling of #2503's `internal/maintenance/jobs` pass.

## The declarations now state what each helper may do

| slice | methods | says |
|---|---|---|
| `bookFileCoreScanner` | 1 (3 users) | reads the book_file table; **cannot write** |
| `bookFileBulkDeleter` | 1 | the destructive half of the missing-file repair; **cannot touch a book row** |
| `fsRegroupStore` | 11 | genuinely moves files between books and deletes rows — now visible at a glance |

Explicit method lists, **not** embedded `database` sub-interfaces: embedding reads tidier but is not narrowing, since `database.BookStore` alone is 51 methods. The measured cost of that mistake is `internal/reconcile`, which composes four sub-interfaces to declare 115 methods and uses 11.

## Verification

- `go build ./...` exit 0; full-tree `go vet ./...` exit 0
- `go test` green on `internal/plugins/maintenance`, `internal/maintenance`, `internal/maintenance/jobs`
- **Monotonicity holds: zero call sites and zero test files changed.** 19 signature lines, 19 insertions, 19 deletions across 13 files. Diff is exactly 41 non-header lines = 19 signature pairs + 1 removed import + the count fix.
- The 15 compile-time assertions were verified to build *before* any signature was touched, isolating signature typos from parameter edits.

## Two counts corrected

- `jobs/store_slices.go` said **31** `MaintenanceJob.Run` implementers. It is **37** — 37 `Run` receivers, 37 files, 37 non-test `maintenance.Register` calls, one each.
- The `todo.d` fragment said **24** store-parameter declarations remain. Re-measured: **54** (8 maintenance `PROP`-tier + 65 outside maintenance, of which 30 have an entirely unused store param and want the parameter deleted rather than narrowed). The old figure undercounted free functions by 3 and covered only the maintenance packages.

## Note on single-method slices

Deliberate. Option D from the audit (pass the method value, no interface) would say the same thing with no type declaration, but it rewrites every call site; a one-method interface keeps them untouched, which is what makes a 19-site sweep reviewable in one pass. D stays available per-site later.


jdfalk:
Closing: **already in `main`**, and a rebase produces an empty branch.

Rebased it onto current `main` as requested — `git rebase` **skipped the commit itself** with `warning: skipped previously applied commit 6a74f929`. Confirmed three ways rather than on the rebase alone:

- **Commit:** `fc861a9c` in `main` is this same change; `6a74f929` here is the pre-rebase SHA. Merged during the GitHub outage via a local rebase-merge, so the merge never went through GitHub's button and the PR stayed open showing `mergeable=false / dirty`.
- **Content:** all **18** touched files are byte-identical to `main` (0 differ). Negative control in the same run — an unmerged branch — correctly reported 44 of 44 files differing, so the check discriminates rather than reporting zero for everything.
- **Artifact:** `internal/plugins/maintenance/store_slices.go` is present on `main` at its full 186 lines with all 15 narrow interface declarations.

The `dirty` mergeable state is real but misleading here: the branch is 18 commits behind `main` and its own patch is already upstream, so there is nothing left to replay. Closing rather than rebasing. Same disposition as #2523/#2524/#2525/#2526.

PR #2523, opened 2026-08-17
refactor(mocks): drop 41 generated database mocks nothing referenced

jdfalk:
## What

Removes **41 of the 45** generated `internal/database` mocks — **24,613 lines, 46.7%** of `mock_store.go`, which goes from **52,753 → 28,139** lines.

Only `.mockery.yaml` entries were edited. The file is generated, so restoring the entries and running `make mocks` reverses this byte-for-byte at the pinned v3.7.1.

## Kept, because they are referenced

| mock | refs | files |
|---|---|---|
| `Store` | 354 | 54 |
| `ImportPathStore` | 6 | 2 |
| `OpsV2Store` | 5 | 2 |
| `OperationStore` | 2 | 1 |

## Why they were dead

Structurally, not accidentally. They mock `database`'s **wide sub-interfaces** — `BookStore` alone is 51 methods — and **no production signature is typed as one**, because constructors take the whole `database.Store`. Generating a mock nothing can use is the predictable outcome.

Narrowing (#2503, #2521) is the other half of that story but **does not resurrect these**: narrow slices are package-local interfaces, so these 41 stay dead either way. Worth stating plainly so the two workstreams don't get credit for each other's wins.

## The census took three passes — please read this bit

1. A **bare-name grep** said 9 were used. 8 of the 45 names are also declared in other packages, so it was counting other packages' types.
2. An **alias-qualified grep** (the package is imported as `mocks`, `dbmocks`, and `databasemocks`) said 3 used across 13 files — impossible against 57 importers. It was blind to mockery's **`NewMockX` constructor form**, which is how 54 of the 57 actually use it.
3. The type-declaration pattern `^type Mock[A-Za-z]+ struct` silently skipped **`MockOpsV2Store`** — the character class excludes the digit in `V2`. **That one has 5 live references and would have been deleted.** It hid because the same blind spot removed it from the numerator *and* the population, so the ratio still read "41 unused" and looked right.

Negative control (`MockTotallyFakeStore`) returns 0 on the final pattern.

## Verification

- `go build ./...` exit 0
- Full-tree `go vet ./...` exit 0 — **vet compiles test files**, so no test references a deleted mock
- `go test ./internal/database/mocks/...` ok — including the reflection-based coverage test that walks all 399 `Store` methods
- `make mocks-check` clean after commit (regeneration is idempotent at v3.7.1)
- The 24,613 deleted lines were uncovered generated code, so `Coverage Floor` should move **up**


jdfalk:
Closing: already in `main`. Verified two ways — `git cherry origin/main` reports 0 new commits, and all 3 touched files are byte-identical to `main`. Merged during the GitHub outage via a local rebase-merge, so the merge never went through GitHub's button and the PR stayed open. Nothing is lost by closing; a rebase would replay nothing.

PR #2524, opened 2026-08-17
docs(plans): measure the scan phase split, plan the AI-pass separation

jdfalk:
Answers the question *"why don't we split those into two tasks so they're not interleaved"* with a measurement rather than an opinion.

## The number

Sampled the running production scan every 10s, classifying each sample by `progress_message` shape. 13.1-minute window, 65 samples:

| phase | wall-clock | share |
|---|---:|---:|
| `ai-parse` | 8.8 min | **66.9%** |
| `metadata` | 4.3 min | 33.1% |

723 books in 13.1 min = 0.92 books/s. Excluding AI time: **2.78 books/s**.

Instantaneous rates within the window vary by more than 10× (4.7 books/s in a metadata stretch, 0.28 across a stretch containing an AI block), so no single average of the scan is meaningful. The plan states the window's scope explicitly rather than quoting 3.0× as a whole-scan speedup.

## Why it matters for checkpointing

The AI phase runs *inside* each 500-book chunk, **before** the chunk-end checkpoint — so ~12 min of AI work per chunk is uncheckpointed and repeated on restart. That makes it a better target for the resume work than the walk phase.

## Why this is a plan and not a patch

AI results are merged into the book **before** its first save. Splitting the phases means the AI pass runs after rows are persisted, so it must *update* them — and a bare whole-row `UpdateBook` is this repo's dominant data-loss shape. The narrow field-level update lands first, alone, with a mutation-tested preservation guarantee.

## The separable win

The batch loop sleeps 2s between batches — a rate-limit courtesy for a hosted, quota-metered API. Production's backend is a **local Ollama** (qwen2.5:7b-instruct) with no quota. A direct probe:

| shape | wall | speedup |
|---|---:|---:|
| serial n=3 | 0.81s | 1.00× |
| concurrent n=3 | 0.44s | 1.86× |
| concurrent n=6 | 0.67s | 2.43× |

It does not serialize. **But that probe used 8-token completions at 0.27s each, against real batches of ~21s** — opposite ends of the prefill/decode curve. The plan says explicitly not to carry 2.43× to production and to re-probe with real batches before sizing a pool.

Docs only. No code change.

### DIFF
--- a/changelog.d/20260817_160000_docs_scan_ai_phase_plan.md
+++ b/changelog.d/20260817_160000_docs_scan_ai_phase_plan.md
@@ -1,0 +1,21 @@
+### Added
+
+- **Plan: split the scan's AI-parsing phase out of the metadata pass**
+  (`docs/plans/2026-08-17-split-scan-ai-phase.md`). Measured on the running
+  production scan, AI parsing consumes **66.9% of scan wall-clock** while the
+  metadata pass consumes 33.1%; excluding AI time moves throughput from 0.92 to
+  2.78 books/s over the sampled window.
+
+  The AI phase is also the phase the checkpoint cannot see — it runs inside each
+  500-book chunk, before the chunk-end checkpoint, so ~12 minutes of work per
+  chunk is repeated on restart.
+
+  The plan records a hazard that keeps it a plan: AI results are merged into the
+  book *before* it is first saved, so splitting the phases means the AI pass must
+  update already-persisted rows — and a bare whole-row update is this repo's
+  dominant data-loss shape. A narrow field-level update lands first, alone.
+
+  One finding is separable and carries none of that risk: the batch loop sleeps
+  2s between batches as a rate-limit courtesy, but production's LLM backend is a
+  local Ollama with no quota, and a direct probe shows it serves concurrent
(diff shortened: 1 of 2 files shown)


jdfalk:
Closing: **superseded, and merging this would regress `main`.**

`git cherry origin/main` reports 0 new commits, but `docs/plans/2026-08-17-split-scan-ai-phase.md` still differs — this branch holds **v1.1.0** while `main` has **v1.2.0**. The branch is the *older* revision, so merging or rebase-merging it would revert `main`'s newer content.

What it would revert is not cosmetic. `main`'s v1.2.0 (commit `3ede9161`) **retracted the recommendation to give the AI batch loop a bounded worker pool**, on direct GPU measurement: the LLM host was at 97 °C — above its own 95 °C shutdown spec — with hardware thermal slowdown active and clocks pinned to 87%, i.e. saturated rather than idle. The original recommendation rested on a probe using 8-token completions of a trivial prompt at ~0.27s each, while a real batch is 20 filenames taking ~21s; those sit at opposite ends of the prefill/decode curve, so the probe measured request-latency headroom, not compute headroom.

This branch still contains that withdrawn recommendation and its superseded probe table. Closing rather than rebasing.

PR #2525, opened 2026-08-17
fix(abs): register HEAD for the byte-serving routes

jdfalk:
gin registers one method per call, so a GET-only route answers `HEAD` with **404, not 405** — indistinguishable from "this file does not exist."

## Why it matters

That silently destroyed a real measurement. A probe built to find `book_file` rows with no bytes behind them reported **100% of 1,786 files missing** and passed its own sanity check, because a fabricated ino returned the same 404 as every real one. A uniformly-dead instrument agrees with any hypothesis.

Verified against production before the change, on the **same URL**:

| request | status |
|---|---|
| `GET /api/items/<real>` | **200** |
| `HEAD /api/items/<real>` | **404** |
| `GET /api/items/<bogus>` | 404 |
| `HEAD /api/items/<bogus>` | 404 |

(My first probe was itself uniformly dead — `GET` on a real *ino* also 404s, because that item's files are among the 41.8% with no bytes. The 200 above is from an endpoint that genuinely serves, which is what makes the comparison mean anything.)

## What changed

Three route registrations. **No handler changed.** All three end in `ServeFileWithRange` → `http.ServeContent`, which omits the body for HEAD while still writing `Content-Length`, `Content-Type`, `ETag` and `Accept-Ranges`. `ABSRequireAuth` already accepted `?token=` on `MethodHead` (`middleware/absauth.go:578,603`) — someone anticipated HEAD; only the routing table was missing.

## Verification

Full `abs` package green (30.7s), `go vet` clean. Every registration mutation-tested:

| Mutation | Result |
|---|---|
| unregister `HEAD /file/:ino` | ✅ killed |
| unregister `HEAD /file/:ino/download` | ✅ killed |
| unregister `HEAD /cover` | ❌ **survived** → fixed |

**The cover mutation surviving is the finding worth keeping.** That test asserted only "HEAD and GET agree" — and with no cover in the fixture both answered 404, so it agreed while both were dead. The test for uniformly-dead instruments was itself a uniformly-dead instrument. Seeding a real cover makes GET 200, the known-good half is now asserted first and fatally, and the mutation is killed.

Every test here asserts a known-good **and** a known-bad in the same run, so a router regressing to 404-for-everything fails the good half instead of going quietly green.

### DIFF
--- a/changelog.d/20260817_170000_fix_head_file_routes.md
+++ b/changelog.d/20260817_170000_fix_head_file_routes.md
@@ -1,0 +1,16 @@
+### Fixed
+
+- **`HEAD` now works on the ABS byte-serving routes** — `/api/items/:id/file/:ino`,
+  `.../download`, and `/api/items/:id/cover`. gin registers one method per call,
+  so a GET-only route answered `HEAD` with **404, not 405**, which is
+  indistinguishable from "this file does not exist".
+
+  That silently destroyed a real measurement. A probe built to find `book_file`
+  rows with no bytes behind them reported **100% of 1,786 files missing** and
+  passed its own sanity check, because a fabricated file id returned the same 404
+  as every real one. A uniformly-dead instrument agrees with any hypothesis.
+
+  No handler changed. These routes already end in `http.ServeContent`, which
+  omits the body for `HEAD` while still sending `Content-Length`, `Content-Type`,
+  `ETag` and `Accept-Ranges`, and the ABS auth middleware already accepted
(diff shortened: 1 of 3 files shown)


jdfalk:
Closing: already in `main`. Verified two ways — `git cherry origin/main` reports 0 new commits, and all 3 touched files are byte-identical to `main`. Merged during the GitHub outage via a local rebase-merge, so the merge never went through GitHub's button and the PR stayed open. Nothing is lost by closing; a rebase would replay nothing.

PR #2526, opened 2026-08-17
docs(audit): root-cause the orphan destination rows

jdfalk:
Answers "why did the organizer record destination rows it never populated?" — the repair op deletes those rows; this is about what creates them.

## The mechanism

`resolveOrganizedFilePath` (`internal/organizer/service.go:1254`) has three branches, and its own comment states the third plainly:

> prefer the target if it exists, fall back to the source if THAT still exists, **and only then take the plan on faith**

So when **neither** the planned target **nor** the source exists on disk, it returns the planned path — and that value goes straight into `CreateBookFile` at `service.go:1444`. That is exactly a destination row recorded for bytes never written.

Two properties match the observed shape:

- A planned path is always `RootDir` + the naming patterns, so **every row created this way is under the organizer's tree and none can be under iTunes** — which is the reported distribution exactly.
- The third branch is **silent**. The source-exists branch logs a warning; this one does not. Verified by reading all three.

The single-file branch just below (`else if !isDir { newBF.FilePath = newPath }`) has no disk check at all.

## Why #2479 is the suspect for bulk

`resolveOrganizedFilePath` *recomputes* a plan for an organize that already ran, and its comment names the fragility: "only correct while both calls see the same inputs." #2479 changed those inputs — the file naming pattern now decides organized filenames, where before `filepath.Base(src)` was kept. So for a multi-file book organized before #2479: bytes live under the old name, the source was consumed by that move, the recomputed plan names a pattern-derived sibling that never existed → target missing, source missing → third branch.

The file is sitting in the same directory under its old name. The row names a sibling that does not exist.

## What this does NOT claim

**Not** that this produced the 41.8%. That figure is extrapolated from 120 books; this is a mechanism, not a measured population. The note says so explicitly and does not date the rows.

## 🔴 This may change the approved repair

`missing-file-repair` **deletes** dead rows. If the bytes are present under a different filename, deleting is the **wrong** action — it discards the only pointer to a file that still exists. The note carries a directory-listing test that discriminates the three cases, and recommends running it before `--apply`.

Docs only. No code change.

### DIFF
--- a/changelog.d/20260817_180000_docs_orphan_rows_root_cause.md
+++ b/changelog.d/20260817_180000_docs_orphan_rows_root_cause.md
@@ -1,0 +1,20 @@
+### Added
+
+- **Root-cause note for the orphan destination rows**
+  (`docs/audits/2026-08-17-orphan-destination-rows-root-cause.md`). Identifies the
+  code path that records a `book_file` row for bytes that were never written:
+  `resolveOrganizedFilePath` (`internal/organizer/service.go:1254`) prefers the
+  planned target if it exists, falls back to the source if that exists, and
+  **when neither exists returns the planned path anyway** — silently, with no
+  warning on that branch. The value goes straight into `CreateBookFile`.
+
+  Because a planned path is always built from `RootDir` plus the naming
+  patterns, every row created this way lands under the organizer's own tree and
+  none can land under iTunes — which matches the observed shape exactly.
+
+  🔴 **This may change the repair.** #2479 made the file-naming pattern decide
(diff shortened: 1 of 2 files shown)


jdfalk:
Closing: already in `main`. Verified two ways — `git cherry origin/main` reports 0 new commits, and both touched files are byte-identical to `main`. Merged during the GitHub outage via a local rebase-merge, so the merge never went through GitHub's button and the PR stayed open. Nothing is lost by closing; a rebase would replay nothing.

